PR #5, opened 2015-04-16
baker: dump logfiles on stdout on ERRORs

kimrhh:
This is very useful when working with buildbot, to avoid having to inspect logfiles on the slave machines.


esben:
Good idea, especially in combination with CI.  But....  when doing normal developer builds, dumping multiple potentially huge logfiles on console might not be the best idea, especially not when working over a slow ssh connection.

A command-line option for getting this behaviour should be added.

esben:
Or perhaps just dump the last 42 lines by default, allowing a command-line option of overriding that value, with -1 for all lines.

esben:
I created a new commit with the requested changes.  Please review.

PR #6, opened 2015-04-17
Toolchain upgrade

esben:
Upgrade toolchain recipes.

binutils 2.25
glibc 2.21
gcc 4.9.2
libtool 2.4.6


esben:
Only remaining build issue is with valgrind, which should be fixed with oe-lite/base#17 which is pending merge of this PR (chicken and egg problem).

So let's merge this PR first.

PR #13, opened 2015-05-22
Improvements

diorcety:


recipes/libtool/libtool-2.4.2.inc
```diff
@@ -0,0 +1,14 @@
+LICENSE = "GPL-2.0+"
+LICENSE_${PN}-libltdl = "LGPL-2.1+"
+
+require ${PN}.inc
+
+SRC_URI:>native = " file://syslib.patch"
+SRC_URI:>cross = " file://syslib.patch"
+SRC_URI:>sdk-cross = " file://syslib.patch"
+SRC_URI:>cross = "file://cross_compile.patch"
```
esben: Missing a space before file://

recipes/libtool/libtool-2.4.2.inc
```diff
@@ -0,0 +1,14 @@
...
+SRC_URI:>sdk-cross = "file://cross_compile.patch"
```
esben: Missing a space before file://

recipes/libtool/libtool_2.4.2.oe
```diff
@@ -1,19 +1,12 @@
...
-addtask bootstrap before configure after fetch unpack patch stage
-do_bootstrap[dirs] = "${S}"
+RECIPE_TYPES = "machine cross sdk-cross"
 
-do_bootstrap () {
-	if [ "${RECIPE_TYPE}" != "native" ]; then
-		./bootstrap
-	fi
-}
+DEPENDS += "native:help2man"
```
esben: I guess this is correct, but...

OE-lite/core recipes must not depend on anything from OE-lite/base.  It must be possible to build OE-lite/core by itself.  In this case, the solution could be to convince libtool to not build man-pages (by default), and thus avoid this dependency.

recipes/libtool/libtool-2.4.2.inc
diorcety: Indeed ... not discover cause that multiple appends on same flag and variable are not handled (use only the last one) which is on this case SRC_URI:>cross = " ${SRC_URI_WINDOWS}"

Btw the missing  patches break some x11 libs compilation

recipes/libtool/libtool_2.4.2.oe
diorcety: we can put help2man in core, it's good for me

esben: Will you update this PR with a commit adding help2man to OE-lite/core?

diorcety: Ok i will do a PR for that :)

classes/autoconf.oeclass
```diff
@@ -0,0 +1,24 @@
+
```
esben: Could we rename this to automate.oeclass ?

classes/gettext.oeclass
```diff
@@ -5,9 +5,12 @@
 ##
 ## @useflag nls Enable nls
 
-CLASS_DEPENDS =+ "${DEPENDS_GETTEXT}"
-DEPENDS_GETTEXT = ""
-DEPENDS_GETTEXT:USE_nls = "gettext native:gettext-utils libiconv libintl"
+CLASS_DEPENDS =+ "${DEPENDS_GETTEXT_NLS} ${DEPENDS_GETTEXT_NATIVE}"
+DEPENDS_GETTEXT_NATIVE = ""
+DEPENDS_GETTEXT_NATIVE:native = "native:gettext-utils"
```
esben: So if you have USE_nls flag disabled, you need native:gettext-utils for native recipes, but nothing for other recipe types?  Why/how is that?

classes/stage.oeclass
```diff
@@ -6,6 +6,7 @@
 ## them.
 
 inherit binconfig-stage
+inherit wrapper-stage
```
esben: You have already moved this commit to PR #14, so please drop it from this PR.

classes/wrapper-stage.oeclass
```diff
@@ -0,0 +1,60 @@
+## Class for packages depending on recipes having wrapper files.
```
esben: You have already moved this commit to PR #14, so please drop it from this PR.

classes/wrapper.oeclass
```diff
@@ -0,0 +1,55 @@
+## Class for packages having wrapper files.
```
esben: You have already moved this commit to PR #14, so please drop it from this PR.

recipes/binutils/binutils.inc
```diff
@@ -129,9 +130,6 @@ RDEPENDS_${PN}-strip = "libbfd"
 RDEPENDS_${PN}-ar = "libbfd"
 RDEPENDS_${PN}-strings = "libbfd"
 
-# The binutils package is used to pull in all provided util features
-RDEPENDS_${PN} = "${AUTO_PACKAGE_UTILS_PROVIDES}"
```
esben: You mentioned earlier that this line was redundant.  How is it that?  How is all the auto-package-utils packages pulled in by RDEPENDS_${PN} without this line?

recipes/eglibc/eglibc_2.18-r24943.oe.sig
```diff
@@ -1 +1,2 @@
+517cfd34b8d250dcdca0f3490fa4e9bf571ee473  svn://www.eglibc.org/branches/eglibc-2_18;rev=r24943
```
esben: As mentioned, we should investigate why this signature is different.  I have a copy of the old ingredient file, so I guess I should investigate it.

recipes/gcc/gcc.inc
```diff
@@ -48,6 +48,7 @@ PROVIDES_${PN}[qa] += "allow-missing-provides:liblto-plugin"
 # Get rid of gcc-dev package
 PACKAGES = "${PN}-dbg ${PN}-doc ${PN}-locale ${PN}"
 FILES_${PN} += "${libdir}/*.a"
+FILES_${PN} += "${libdir}64/*.a"
```
esben: Please move this commit to separate commit (together with the similar binutils commit).

We need to investigate what is going on here, but let's avoid blocking the rest of this PR by that.

recipes/libtool/libtool_2.4.2.oe
esben: Or better, add that commit to this PR, as it is needed by this change.

classes/gettext.oeclass
diorcety: some native tools need gettext in order to be compiled

recipes/binutils/binutils.inc
diorcety: Please look the last 2 lines of this recipe

esben: Ah, I see.  Sorry about that :)

recipes/libtool/libtool_2.4.2.oe
esben: help2man is now in OE-lite/core.

classes/gettext.oeclass
esben: And gettext is not needed when cross-building the same tools?

esben:
@diorcety I would really like to merge as much as possible of this PR before 5.0.0.  Could you update this commit according to the various discussions above?

classes/gettext.oeclass
diorcety: Not sure on this point... I have to give a boost on my oe-lite PR :)

esben:
As the wrapper commit (bf420e7fade1aa3859c8911999ab779b4d0a7e54) has already been merged,
and most of the other commits are now included in PR #89, the only remaining commits in this PR is:

* The binutils and gcc lib64 commits: d5b775907240dfbef11fc15cfa8587d3b2a4d4f6 and 3916e2aff86a9fa7fc961ec83e683be0edf6c494.
* The libtool commit: 6832cc9cf578c7c2ba4a4f8bd82f76b283e5fbc1

The libtool commit need to be ported the the current libtool version (2.4.6).

The two lib64 would be good to have investigated.  I really don't think it is right to include them, but we clearly need to work out the problem you have experienced.  Maybe it is related to the problems @sknsean and @kimrhh are seeing in PR #64.

esben:
Hmm... The automate commit (aa47ed43181319ac8baacbca7532d978f415482f) causes quite some
build problems with current OE-lite/base, so I am leaving that out of PR #89 also, so a separate PR for that is needed also.

esben:
This is getting really old.  A lot of the stuff has been merged in separate PR's.  Anything remaining that could/should be merged, please create new PR(s).

PR #16, opened 2015-05-25
eglibc: use svn protocol instead of http (issue with web server)

diorcety:


recipes/eglibc/eglibc_2.18-r24943.oe.sig
```diff
@@ -1 +1,2 @@
+517cfd34b8d250dcdca0f3490fa4e9bf571ee473  svn://www.eglibc.org/branches/eglibc-2_18;rev=r24943
```
esben: Hmm...  I wonder why the signatures are different.  Hopefully this is only timestamp that differs.

esben: Wow, this is odd.

When checking out with svn protocol, you get one additional file compared to checkout with http:
linuxthreads/linuxthreads/manager.c

Any ideas for how this makes sense?

esben: Based on the svn log information, I would say that the checkout with svn protocol (ie. with this PR) is correct, and the old result was wrong.

PR #19, opened 2015-06-02
Add help2man recipe

diorcety:


help2man/help2man.inc
```diff
@@ -0,0 +1,9 @@
+# -*- mode:python; -*-
+RECIPE_TYPES = "machine native"
+
+EXTRA_AUTORECONF="-Iaclocal.m4"
+inherit autotools-autoreconf
```
esben: Is it necessary to do an autoreconf?

help2man/help2man.inc
```diff
@@ -0,0 +1,9 @@
+# -*- mode:python; -*-
+RECIPE_TYPES = "machine native"
```
esben: Any special reason for RECIPE_TYPE=machine ?

diorcety: have help2man in the image? Yes not so usefull :)

esben: Hehe, let's just change this to RECIPE_TYPES = "native" then :)  That way, we don't have to waste buildbot time building it for all machine archs.

help2man/help2man.inc
esben: Did you try just using "inherit autotools", and did that cause any problems?

diorcety: can't tell i have to test in on my fedora, centos 6 and ubuntu in order to see if the tools versions are good enough

esben: Did you find time to test if we could stay with just autotools instead of autotools-autoreconf?

PR #22, opened 2015-06-12
Kbuild class

esben:


kimrhh:
maybe make it do some of the stuff done in the kbuild use cases?

like e.g.

EXTRA_OEMAKE += "ARCH=${TARGET_ARCH} CROSS_COMPILE=${TARGET_PREFIX}"
EXTRA_OEMAKE += "CROSS=${HOST_PREFIX}"

esben:
How much of that is kbuild command, and not specific to a project using kbuild (like Linux kernel)?

Setting both CROSS and CROSS_COMPILE is suspicious...

esben:
It might make sense to add ARCH and CROSS_COMPILE to EXTRA_OEMAKE, but that can be done later.  Would be nice to checkup on status of this, if it is really safe to do that for all kbuild based projects.

PR #23, opened 2015-06-15
git fetcher: avoid mirror of local git repositories

kimrhh:


lib/oelite/fetch/git.py
```diff
@@ -267,6 +267,17 @@ def unpack(self, d):
         return True
 
     def mirror(self, mirror=os.getcwd()):
+        if self.is_local:
+            # Don't mirror if the source is local. Concept taken from fetch
+            # Do check revision exists before accepting though
+            print "Warning: Skipping creating mirror for local path %s"%(self.uri)
```
esben: Does this really have to be a warning?  Isn't it more a info/debug message?

lib/oelite/fetch/git.py
```diff
@@ -267,6 +267,17 @@ def unpack(self, d):
...
+            repo = oelite.git.GitRepository(self.repo)
+            if self.has_rev(repo):
```
esben: Why not just return True right away in this case?  Why check for availability of revision when we are not going to do anything with it...

lib/oelite/fetch/git.py
kimrhh: In the line of "least surprises" it might be nice to be told that stuff is not being mirrored, whether it should say warning *shrug* :)

lib/oelite/fetch/git.py
kimrhh: We decided to duplicate the logic from fetch, since we where not sure if you could hit this without running fetch as well. Personally i think the check belongs higher in the hierachy. We can just make it return True if you prefer that.

lib/oelite/fetch/git.py
esben: I agree, it is a good thing to notify the user that stuff is not being mirrored.  But it should generally be possible to avoid warnings, which mandates that this notification is not really warning.

lib/oelite/fetch/git.py
esben: The mirror task is run after fetch, so I don't think this duplication of code makes a lot of sense.  It doesn't hurt a lot either, just a slightly confusing, as it kind of hints that it makes sense to check for it here.  So yes, I prefer to get rid of this code here :)

kimrhh: agreed, makes sense. will change that

lib/oelite/fetch/git.py
kimrhh: I understand your reasoning, but what do you want, no message at all? imo a warning is ok when you are trying to do something contradictory. Here you have instructed oe-lite to mirror something with a local url. Just silently ignoring it would likely end up with someone having to read through the source to figure out what is going on :) But, just make a dictator decision and let me know how you want it :)

esben: Dictator decision: Replace the 'Warning:' prefix with something like 'Info:' instead.

Usecase: You have created a recipe with a local git SRC_URI and execute a mirrorall task. You haven't specifically requested to mirror the local git SRC_URI, and actually have no good way to get rid of that warning while doing the perfectly valid mirrorall task.

esben:
LGTM

esben:
Uh, better remove the 4.0.1 milestone for this pr against master.  It should be used when you create a backport of this to 4.0 branch.

PR #24, opened 2015-06-16
GCC 5

esben:


esben:
The gcc recipe should be upgraded to 5.2.0 in this PR.

sknsean:
Hi
I have updated gcc to 5.2. It's not without problems...
My 3.10-lts fails with:
include/linux/compiler-gcc.h:103:30: fatal error: linux/compiler-gcc5.h: No such file or directory
Ncurses6.0 fails with:
_1167.c:843:15: error: expected ')' before 'int'
../include/curses.h:1631:56: note: in definition of macro 'mouse_trafo'
 #define mouse_trafo(y,x,to_screen) wmouse_trafo(stdscr,y,x,to_screen)
I will look for a workaround :-)
/Sean

esben:
Yes, I ran into the ncurses problem with gcc 5.1 also.  I believe upstream ncurses has some kind of fix for it.

sknsean:
I have found 2 patches for the kernel and tested it...
I also found a patch for ncurses :-)

However gcc complains in configure with:
Switch "--with-tune" may not be used with switch "--with-cpu"...
Is it enough to the cpu type and float? As I understand it the tune and arch follows the cpu type...

A workaround for my arm926ejs, have been to remove the march and mtune from the arch table...
         '926ejs'               : {
-            'march'            : 'armv5te',
+#            'march'           : 'armv5te',
             'mcpu'             : 'arm926ej-s',
-            'mtune'            : 'arm926ej-s',
+#            'mtune'           : 'arm926ej-s',

esben:
GCC manual on -mcpu:
> -mcpu=name
> This specifies the name of the target ARM processor. GCC uses this name to derive the name of the target ARM architecture (as if specified by -march) and the ARM processor type for which to tune for performance (as if specified by -mtune). Where this option is used in conjunction with -march or -mtune, those options take precedence over the appropriate part of this option.

So we should probably either use mcpu OR (march and mtune). I haven't seen any documentation on the fact that we are not allowed to use all 3 at once, but it clearly does not make much sense anyway.

Maybe it would make sense to just use mcpu, and thus allow manual override both with mcpu and march and mtune.

sknsean:
I haven't seen it in any doc either...
But in gcc/config.gcc at line 3721 the are these functions:
if test "x$with_arch" != x && test "x$with_cpu" != x; then
   echo "Switch \"--with-arch\" may not be used with switch \"--with-cpu\""  1>&2
   exit 1
fi

if test "x$with_cpu" != x && test "x$with_tune" != x; then
   echo "Switch \"--with-tune\" may not be used with switch \"--with-cpu\""  1>&2
   exit 1
fi

esben:
Superseded by PR #66

Issue #27, opened 2015-06-29
ply fails on python-2.6


hundeboll:
Running OE-lite on an ancient Redhat machine generates the following error:

```
ERROR: exception in bake.run()

Traceback (most recent call last):
  File "/mnt/tmp/buildbot/slave/maquet-deploy/build/bakery/oebakery/cmd/cmds.py", line 137, in call
    ret = function(*args)
  File "/mnt/tmp/buildbot/slave/maquet-deploy/build/oe-lite/meta/core/lib/oelite/cmd/bake.py", line 29, in run
    baker = oelite.baker.OEliteBaker(options, args, config)
  File "/mnt/tmp/buildbot/slave/maquet-deploy/build/oe-lite/meta/core/lib/oelite/baker.py", line 115, in __init__
    self.confparser = confparse.ConfParser(self.config)
  File "/mnt/tmp/buildbot/slave/maquet-deploy/build/oe-lite/meta/core/lib/oelite/parse/confparse.py", line 10, in __init__
    super(ConfParser, self).__init__(data, parent, **kwargs)
  File "/mnt/tmp/buildbot/slave/maquet-deploy/build/oe-lite/meta/core/lib/oelite/parse/oeparse.py", line 18, in __init__
    self.tokens = lexer.lextokens.keys()
AttributeError: 'set' object has no attribute 'keys'

CRITICAL: bake failed: Exception: 'set' object has no attribute 'keys'
```

I am no expert in python-ply, but let me know if I can do any testing.

```
$ python --version
Python 2.6.6
```

hundeboll:
After digging around, it turns out I had ply-3.6 installed, while oelite/core works only with ply-3.4... Downgrading removed the issue.

esben:
Ok.

Although I cannot point at something concrete, I suspect that you might run into some problems using Python 2.6 (or older) with current OE-lite/core.

esben:
Also, FWIW: From changelog of PLY 3.5:

04/16/15: beazley
          Support for versions of Python <2.7 is officially dropped.  PLY may work, but
          the unit tests requires Python 2.7 or newer.

So you might not want to use PLY 3.6 with Python 2.6 anyway.

mikaelj:
Python-2.7 on PLY-3.6.

Which Python versions are supported for PLY-3.6?

kimrhh:
[...]

PR #28, opened 2015-08-12
Issue #27: ply compatibility

mikaelj:
PLY >= 3.6 uses a set instead of a dict for the lextokens. It's just used as a list in the end, so check the type of the return value to determine how to convert the data.


esben:
Looks good... If it wasn't for the removal of the prefer parsing part.  Please update PR, keeping onlyl the first hunk.

esben:
Great.  Would you mind squashing the two commits and force-update your branch?

esben:
I wonder why the kbuild commits turned up in this PR, but they are fine also.  Merging :)

PR #33, opened 2015-08-26
Fix LOADADDR for USE_kernel_uimage_entrypoint

mikaelj:
This fixes the kernel recipe when USE_kernel_uimage_loadaddr is not set and USE_kernel_uimage_entrypoint is set.


esben:
LGTM.

PR #35, opened 2015-09-03
add scons support

hundeboll:


esben:
@hundeboll do you have time to look at my comments (questions)?

esben:
I have merged c override support, so you can use that now (after adding inherit c).

classes/scons.oeclass
```diff
@@ -0,0 +1,28 @@
...
+EXTRA_OESCONS ?= ""
+
+OESCONS = " \
+    --prefix=${D}${prefix} \
+    TARGET_ARCH=${TARGET_ARCH} \
+    TARGET_OS=linux \
+"
+
+OESCONS:>cxx += "CXX=${HOST_CXX}"
+OESCONS:>c += "CC=${HOST_CC}"
```
esben: You need to manually add a space in front of both CXX and CC, as the :> operator does not, and the += only adds a space between the current value (which is fx. OESCONS:>c) and the append value, and not the parent value (ie. OESCONS).

recipes/scons/scons.inc
```diff
@@ -0,0 +1,17 @@
+DESCRIPTION = "A Software Construction Tool"
+LICENSE = "MIT"
+RECIPE_TYPES = "native"
+
+SRC_URI = "http://ftp.osuosl.org/pub/blfs/conglomeration/scons/scons-${PV}.tar.gz"
+
+do_compile() {
+    python setup.py build
```
esben: Does scons have some specific python module requirements?

hundeboll: Not according to either the scons webpage or the Arch Linux package:
http://www.scons.org/doc/production/HTML/scons-user/ch01s03.html
https://www.archlinux.org/packages/extra/any/scons/

PR #48, opened 2015-10-02
gcc: Sets --with-float argument for arm to gcc for selecting default float-abi.

sknsean:


sknsean:
We are building a machine with:
MACHINE_CPU = "arm-imx53"
MACHINE_ABI	= "softfp arm"
This results in arm-cortexa8neonsfp-linux-gnueabi-gcc default is selecting soft as mfloat-abi.
This patch fixes this and default is now softfp.

Is this the right way to do it?

esben:
I think it is the right thing to do...

Could you verify that it also behaves correctly for not softfp arm architectures, fx. arm-926ejs ?

sknsean:
MACHINE_CPU = "arm-at91sam9260" results in:
  -march=                     		armv5te
  -marm                       		[enabled]
  -mfloat-abi=                		soft
  -mfpu=                      		vfp

sknsean:
When we are building a machine with:
MACHINE_CPU = "arm-imx53"
MACHINE_ABI = "softfp arm"
Or
MACHINE_CPU = "arm-imx6"
MACHINE_ABI = "hard"

The default march is armv2 it that a problem?  Or is it the mcpu which determines the architecture?
  -march=                     		armv2
  -mcpu=                      		cortex-a8

esben:
How do you see this armv2 default march?  Sounds like something we want to fix.

Looking good with the arm-at91sam9260 example.  I will be merging moths PR #49 instead of this, as it includes this change, and has a lot of other goodies.

But thanks for the work!

PR #54, opened 2015-10-14
Add use flag to include deprecated rpc functionality

ksorensen:
Add use flag to enable obsolete rcp support

This is necessary in order to be able to use nfs.

Signed-off-by: Klaus Henning Sorensen <klaus.sorensen@prevas.dk>


recipes/eglibc/eglibc.inc
```diff
@@ -24,14 +24,17 @@ LIBTOOL_DEPENDS = ""
 
 MACHINE[unexport]       = "1"
 
-RECIPE_FLAGS:native = "toolchain_kernel_version_native"
-RECIPE_FLAGS:machine = "toolchain_kernel_version_machine"
+RECIPE_FLAGS:native = "toolchain_kernel_version_native eglibc_obsolete_rpc"
```
esben: If we have it for both native and machine, shouldn't we have it for sdk also?

esben: We might also want to call it glibc_obsolete_rpc, preparing for using same flag on newer OE-lite/core.

esben:
Should add the same stuff to our master branch, and thus glibc 2.21?

recipes/eglibc/eglibc.inc
```diff
@@ -24,14 +24,17 @@ LIBTOOL_DEPENDS = ""
...
+RECIPE_FLAGS:machine = "toolchain_kernel_version_machine eglibc_obsolete_rpc"
 RECIPE_FLAGS:sdk = "toolchain_kernel_version_sdk"
 
 DEFAULT_USE_toolchain_kernel_version_native	= "2.6.32"
 DEFAULT_USE_toolchain_kernel_version_machine	= "2.6.32"
 DEFAULT_USE_toolchain_kernel_version_sdk	= "2.6.32"
```
esben: Should we add a
DEFAULT_USE_eglibc_obsolete_rpc = "1"
line, to have this enabled by default?  Or does it add considerably to footprint, so it makes sense to keep it disabled by default?

ksorensen: I cannot image it adds significant footprint. The reason I left it disabled by default is because it is deprecated.

The correct long term solution is to use libtirpc instead of rpc in (e)glibc.

ksorensen: I think Kim H would like the same functionality in the master branch in order to be able to use nfs.

recipes/eglibc/eglibc.inc
ksorensen: It should probably be added for sdk as well.

recipes/eglibc/eglibc.inc
esben: Ok, I think we should consider enabling it by default on the master branch (until we have libtirpc in place).

recipes/eglibc/eglibc.inc
esben: We could also just add --enable-obsolete-rpc for both native and sdk always, as the minimal added size should not matter there.

ksorensen:
Force pushed changes.

Added recipe flag to sdk and enabled as default.

recipes/eglibc/eglibc.inc
```diff
@@ -24,14 +24,18 @@ LIBTOOL_DEPENDS = ""
...
+RECIPE_FLAGS:>machine	= " toolchain_kernel_version_machine eglibc_obsolete_rpc"
+RECIPE_FLAGS:>sdk	= " toolchain_kernel_version_sdk eglibc_obsolete_rpc"
 
 DEFAULT_USE_toolchain_kernel_version_native	= "2.6.32"
 DEFAULT_USE_toolchain_kernel_version_machine	= "2.6.32"
 DEFAULT_USE_toolchain_kernel_version_sdk	= "2.6.32"
 
+OBSOLETE_RPC = ""
+DEFAULT_USE_eglibc_obsolete_rpc = "1"
+OBSOLETE_RPC:>USE_eglibc_obsolete_rpc = " --enable-obsolete-rpc"
```
esben: You could just have used OBSOLETE_RPC:USE_eglibc_obsolete_rpc but the result would have been the same.

ksorensen:
The build fails when building rsync for powerpc

From the log file:

```
+ make -j8
perl ./mkproto.pl ./*.c ./lib/compat.c
sed 's;\@bindir\@;/usr/bin;g' <./rsync-ssl.in >rsync-ssl
sed 's;\@stunnel4\@;stunnel;g' <./stunnel-rsync.in >stunnel-rsync
sed 's;\@bindir\@;/usr/bin;g' <./stunnel-rsyncd.conf.in >stunnel-rsyncd.conf
Copying srcdir rsync.1
cp: cannot stat `./rsync.1': No such file or directory
make: [man-copy] Error 1 (ignored)
Copying srcdir rsyncd.conf.5
cp: cannot stat `./rsyncd.conf.5': No such file or directory
make: [man-copy] Error 1 (ignored)
In file included from ./rounding.c:20:0:
./rsync.h:1003:19: fatal error: proto.h: No such file or directory
 #include "proto.h"
```

The "Copying srcdir rsync.1" output comes from Makefile.in and proto.h is part of the rsync package.

It seems unrelated to the rpc change.

Have you seen this before or do you have any ideas?

esben:
It should be fixed by the fix merged to OE-lite/base 4.0 today.  I will kick the buildbot to rebuild this PR again.

ksorensen:
I have noticed an issue with this.

I set 

```
MACHINE_USE_eglibc_obsolete_rpc = "1"
```

in my configuration file.

This leads to

```
Deleting MACHINE_USE_eglibc_obsolete_rpc variable from sdk:eglibc_2.18
```

when the recipe for eglibc is processed.

For my current test build, I have removed eglibc_obsolete_rpc from RECIPE_FLAGS:>sdk:

```
RECIPE_FLAGS:>native    = " toolchain_kernel_version_native eglibc_obsolete_rpc"
RECIPE_FLAGS:>machine   = " toolchain_kernel_version_machine eglibc_obsolete_rpc
RECIPE_FLAGS:>sdk       = " toolchain_kernel_version_sdk"
```

Which gets rid of the warning, but may not be what I really try to accomplish.

Any suggestions?

esben:
You could remove it from RECIPE_FLAGS:>native also.

It does not seem relevant to either of them.

ksorensen:
I have force pushed a new version of this.

The binaries that are built work for me.

I have notice something odd though.

If eglibc_obsolete_rpc is included in RECIPE_FLAGS:>native, non-working artifacts are built and some artifacts (including eglibc) are placed in

tmp/work/machine/powerpc-e300c3-linux-gnu
and some are placed in
tmp/work/machine/powerpc-e300c3-linux-gnu.focondrm


If eglibc_obsolete_rpc is not included in RECIPE_FLAGS:>native, working artifacts are built, and all build artifacts are placed in

tmp/work/machine/powerpc-e300c3-linux-gnu.focondrm

I do not know why adding a recipe flag for sdk builds causes this difference.

esben:
Sounds strange, and quite disturbing.

But in this case, I cannot see any reason what so ever for include eglibc_obsolete_rpc in RECIPE_FLAGS for native.

Could you take a look at the resulting native:eglibc metadata, ie.

```
oe show native:eglibc
```

and see where the MACHINE related powerpc-e300c3-linux-gnu is coming from?

If it is not there, it might come from some of a native dependency that has been staged for native:eglibc.

Be specially on the outlook for anything that relates to MACHINE* variables and native recipes.
There is no, absolutely no, good reasons for MACHINE* variables to exist in native recipe metadata.

PR #55, opened 2015-10-14
4.0 powerpc32

ksorensen:
Various fixes for powerpc32 build.

The circular dependency and gettimeofday commits are in glibc-2.19 and not necessary for the master branch.

The 32 bit addressing patch may be relevant for master in some form. It can be improved

  - Maybe use flag to enable, there is some performance impact since all calls resolved by the dynamic linker will be 32 bit.

  - The toolchain should probably use -fPIC as default if 32 bit addressing is supported


esben:
Looks good. If you would split the the 32 bit addressing patch in a separate PR, I would like to merge the two other as they are.

As for the 32 bit addressing patch, I agree, it should be controllable with a USE flag.  And it needs to be applied to master branch also.  This will be easier to handle in a separate PR.

ksorensen:
I'll split this pull request up into three separate pull request

PR #60, opened 2015-10-20
largefile: enforce largefile autotools flags

kimrhh:
Make sure we enable largefile support when the useflag is enabled
by forcing largefile specific autotools flags to the correct settings.

without this e.g. libsndfile would not be built with largefile
support even though the configure switch was on.

### DIFF
--- a/classes/largefile.oeclass
+++ b/classes/largefile.oeclass
@@ -7,7 +7,12 @@
 
 CLASS_FLAGS += "largefile"
 EXTRA_OECONF_LARGEFILE = "--disable-largefile"
-EXTRA_OECONF_LARGEFILE:USE_largefile = "--enable-largefile"
+EXTRA_OECONF_LARGEFILE:USE_largefile = "\
+    --enable-largefile \
+    ac_cv_sys_largefile_source=1 \
+    ac_cv_sys_file_offset_bits=64 \
+    ac_cv_sizeof_off_t=8 \
+"
 EXTRA_OECONF += "${EXTRA_OECONF_LARGEFILE}"
 
 # Local Variables:


kimrhh:
i have not verified for all archs no, but if we stick it there, don't we risk using largefile settings even when that is disabled? I guess I should see how that behaves...

esben:
ac_cv_sizeof_off_t is not a largefile settings, it is just the length in bytes of the off_t type, and should be handled in the same way as all the other ac_cv_sizeof_* variables in fx. site/arm.

kimrhh:
setting it do 8 bytes seems very much like something that should only be done with large file support though. The default will be 4 bytes for a 32bit system, and I can't really grasp what that could happen if that is set when the system is otherwise configured to not support largefile

kimrhh:
As the "GNU C Library Reference Manual" says

off_t
    This is a signed integer type used to represent file sizes. 
    In the GNU C Library, this type is no narrower than int.
    If the source is compiled with _FILE_OFFSET_BITS == 64 this 
    type is transparently replaced by off64_t.
and

off64_t
    This type is used similar to off_t. The difference is that 
    even on 32 bit machines, where the off_t type would have 32 bits,
    off64_t has 64 bits and so is able to address files up to 2^63 bytes
    in length. When compiling with _FILE_OFFSET_BITS == 64 this type 
    is available under the name off_t.

kimrhh:
guess what really should happens should be that autotools checks how libc was built instead of relying on hardcoded settings ?

esben:
Thought experiment:

1. glibc built with largefile support, and thus off_t being replaced by off64_t.
2. recipe built without largefile.oeclass, and possibly using ac_cv_sizeof_off_t=4
3. Here be dragons?

So, putting it in largefile.oeclass is probably not good enough.
Sticking it directly in site/arm is wrong and bad.

kimrhh:
yeah, this is a mess, this is also what people say other places mentioning it. It might be better to add to CFLAGS/LDFLAGS in largefile.oeclass instead, since that would then only be forced on when you want largefile support?

I _think_ the flags are the same for all archs, but let me check that

esben:
If we do it like that (force it to 8 bytes in largefile.oeclass), you still have potential hidden bugs in recipes using ac_cv_sizeof_off_t but not inheriting largefile.oeclass.

What we really should do, is to have glibc provide it's own siteinfo snippet file, which then get's staged in all recipes using glibc, and have siteinfo.oeclass include that when generating siteinfo files.

From current siteinfo.oeclass:

```
    # TODO: could be extended with searching in stage dir, so build
    # dependencies could provide sitefiles instead of piling everything
    # into common files.  When building for MACHINE_ARCH, search for
    # sitefiles in stage/machine/usr/share/config.site/* and each build#
    # dependency should then install their files into it's own config.site
    # subdir.
```

esben:
A similar PR (#61) was merged to master, and will be included in 5.0.0.

It does not seem like a good candidate for being most backwards compatible bugfix, so could we perhaps close this PR?

kimrhh:
just stumbled over this, yes let us close it :)

PR #63, opened 2015-10-22
64bit sdk fix, kernel newer than 4.0 fix, arch.py cleanup

sknsean:
Some fixes and some cleanup

### DIFF
--- a/classes/kernel.oeclass
+++ b/classes/kernel.oeclass
@@ -134,11 +134,11 @@ do_compile_kernel () {
 		export KBUILD_BUILD_VERSION="$BUILD_TAG"
 	fi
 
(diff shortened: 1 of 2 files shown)


esben:
Ok, so any 2.6 kernels and older, and any 3.x kernels older than 3.7.
So if the header layout change went in at 3.7, it should be perfect :)

sknsean:
I haven't checked if the header change was in 3.7 but I guess the original author have done this ;-)

sknsean:
Should I remove the glibc 64bit sdk fix from this pull request?

esben:
Yes, if you remove that, we can merge the other two now. But please repost the 64 bit sdk fix in another PR. We clearly want it merged :)

PR #64, opened 2015-10-23
glibc: fix 64bit sdk build

sknsean:
without this, the libdir related variables ends up being set to
/../lib resulting in installation outside the install dir.

### DIFF
--- a/recipes/glibc/glibc.inc
+++ b/recipes/glibc/glibc.inc
@@ -43,6 +43,7 @@ do_configure_siteinfo() {
 }
 
 EXTRA_OECONF += "\
+	libc_cv_slibdir='/lib' \
 	--disable-profile \
 	--with-tls \
 	--with-headers=${HOST_SYSROOT}${includedir} \


esben:
This is the commit removed from PR #63.

esben:
In aclocal.m4:
```
dnl Default to slibdir named SLIBDIR instead of "lib", and rtlddir
dnl named RTLDDIR instead of "lib".  This is used to put 64-bit
dnl libraries in /lib64.
dnl LIBC_SLIBDIR_RTLDDIR([slibdir], [rtlddir])
AC_DEFUN([LIBC_SLIBDIR_RTLDDIR],
[test -n "$libc_cv_slibdir" ||
case "$prefix" in
/usr | /usr/)
  libc_cv_slibdir=/$1
  libc_cv_rtlddir=/$2
  if test "$libdir" = '${exec_prefix}/lib'; then
    libdir='${exec_prefix}/$1';
    # Locale data can be shared between 32-bit and 64-bit libraries.
    libc_cv_localedir='${exec_prefix}/lib/locale'
  fi
  ;;
esac])
```
and in sysdeps/gnu/configure.ac:
```
LIBC_SLIBDIR_RTLDDIR([lib], [lib])
```

So, if you manually set libc_cv_slibdir but not libc_cv_rtlddir and libc_cv_localedir, it would seem you are in for some trouble.

So maybe adding
```
    libc_cv_rtlddir='${base_libdir}' \
    libc_cv_localedir='${libdir}/locale' \
```

is needed also.

sknsean:
added :-)

esben:
Looks fine.

Scratching my memory...  Doesn't this entire fix overlap with
https://github.com/oe-lite/core/blob/master/recipes/glibc/glibc-2.21/multilib-paths.patch
?

And if so, is the old multilib-patch broken now?

sknsean:
I think the multi-patch is broken... If i'm building an x86_64 sdk the install happens in /opt/pil/oe-lite/tmp/work/sdk/x86_64-sdk_unknown-linux-gnu/glibc-2.21/ not in /opt/pil/oe-lite/tmp/work/sdk/x86_64-sdk_unknown-linux-gnu/glibc-2.21/install

sknsean:
I have tried to remove the multilib-patch.patch and it seems to work :-)

esben:
Well, I wan't to see a green build icon here before merging :)

sknsean:
Well its not glibc's fault libtiff fails on mingw ;-)

esben:
You are probably right, but I still won't merge until I get a working build.  I will try to find out what is causing tiff to fail on mingw32.

esben:
That said, I generally do not feel very confortable with merging this.  Two reasons:

1. I don't remember seeing anything concrete about this build problem with 64 bit sdk. Is it perhaps related to a specific host distro?
2. I remember that this patch was needed, and that it not only solved a real issue, but seemed like the most right thing to do.

So please, could you describe in more details exactly what goes wrong, and how to reproduce the problem.  I suspect it could be related to some host distro that reports something b0rken (or at least unexpected) when calling "gcc -print-multi-os-directory".

On my (Exherbo) distro, I get 
```
$ gcc  -print-multi-os-directory
.
```

On a random Debian 7 host, I get
```
$ gcc  -print-multi-os-directory
../lib
```

kimrhh:
using the current toolchain and trying to build a 64bit SDK currently breaks, because eglibc ends up being installed outside the install directory as Sean already described. We can dig out a log if you want it :) It is however only tested on ubuntu, so you might be on to something regarding the multilib dir. The cause of the problem was found to be a "../lib" in the wrong place, so sound plausible.

That being said, why do we carry a patch for a critical component of oe-lite, that we do not know why we have? Sounds rather dangerous

esben:
Well, I don't think it is fair to say that "we do not know why we have". The reason for this patch is (or possibly was) problems with x86_64 builds, where libraries were placed in /lib64 and /usr/lib64. The fact that this patch seems to be causing problems now on a specific host distro, is IMHO not enough to blindly just drop the patch.  We need to understand what is going on, so we don't just reintroduce problems on hosts other than the one that you are currently using.

kimrhh:
would like to move on with this, as it is causing problems for people building with clean oe-lite.org meta layers.

When I said "don't know why we have" I was referring to your earlier statement about not remembering exactly why it was there, but no you remembered :) We really need to document that though, so let us add that in a patch header. 

Next question: since building without the multilib patch on Debian derivatives seems to work ok, we need to figure out how to reproduce the original issue introducing the multilib patch. Or we should go further with the cv_lib settings, but that seems like hacking even more to me. Can you remember what build host had it failing so we can try to reproduce the original issue?

Any other suggestions for a way forward?

esben:
Hmm... The problems is likely to be that we are using the output from `$CC -print-multi-os-directory`, which we take special care of in the gcc we build, but when using native gcc, it b0rks out because of something it returns.

I just tested in a fresh ubuntu:14.04 docker container (x86_64)

```
root@1f569b689da7:/# gcc -print-multi-os-directory
../lib
```

What does `gcc -print-multi-os-directory` return on a PC where you observe this problem?

sknsean:
skn@e3cdedf487b5:/opt/pil/oe-lite/src/linux$ gcc -print-multi-os-directory                                                              
../lib
skn@e3cdedf487b5:/opt/pil/oe-lite/src/linux$ arm-cortexa9neon-linux-gnueabi-gcc -print-multi-os-directory                               
.
Thats with ``` EXTRA_OECONF += "\

    libc_cv_slibdir='/lib' ``` in glibc.inc

sknsean:
Work on this PR will continue in #104

PR #66, opened 2015-10-26
gcc5.3

sknsean:
Take a look :-)


sknsean:
https://github.com/sknsean/core/commit/6c21aadae82a7bd2f392340685201e0ea80a705a

recipes/gcc/gcc-5.2.0.inc
```diff
@@ -0,0 +1,4 @@
+SRC_URI:>TARGET_CPU_x86_64 = " file://x86_64-multilib-dirnames.patch"
+
+SRC_URI += "file://mingw32-path-fixup.patch"
+#SRC_URI += "file://pr-64839-libsanitizer.patch"
```
esben: You should probably remove this line.

recipes/gcc/gcc-common.inc
```diff
@@ -137,13 +137,15 @@ do_configure_args ?= ""
...
     args = []
-    for arg_name in ('arch', 'cpu', 'tune'):
+    for arg_name in ('cpu', 'arch', 'tune'):
         if not d.get('ARCH_HAS_WITH_' + arg_name.upper()):
             continue
         arg_value = d.get('TARGET_M' + arg_name.upper())
         if not arg_value:
             continue
         args.append('--with-%s=%s'%(arg_name, arg_value))
+        if "cpu" in arg_name and "arm" in arg_value:
```
esben: Could you add a comment on what this line achieves, so that the when the next guy comes around here doesn't end up with to many question marks?

sknsean:
Fixed the whitespace bug in the comment line :-)

recipes/gcc/gcc-common.inc
```diff
@@ -137,13 +137,16 @@ do_configure_args ?= ""
...
     args = []
-    for arg_name in ('arch', 'cpu', 'tune'):
+    for arg_name in ('cpu', 'arch', 'tune'):
         if not d.get('ARCH_HAS_WITH_' + arg_name.upper()):
             continue
         arg_value = d.get('TARGET_M' + arg_name.upper())
         if not arg_value:
             continue
         args.append('--with-%s=%s'%(arg_name, arg_value))
+	#For arm targets it's only allowed to set either mcpu or march and mtune
```
esben: I still see a tab...

sknsean:
Now i have added the fixup of my whitespace bug and pushed it

esben:
LGTM.  I need to branch out 5.x before merging this though.

esben:
Maybe update to gcc 5.3...

sknsean:
Yes I'm testing with gcc 5.3 :-)

sknsean:
@esben i think this is ready

sknsean:
How about this then?
if "cpu" == arg_name and (arg_value.startswith("arm") or arg_value.startswith("cortex")): 

The cpu arch for arm could be arm11** or cortex-a* therefore this hack :-)

esben:
Uh, I really don't think that CPU arch can ever be cortex-a*.  At least in OE-lite, I believe that we require arm arch tuples must start with arm-*.  Where did you see an cortex-a* arch tuple?

sknsean:
The "cpu" in this context is the mcpu from the arch.py file and therefore yes the cpu can be a cortex-a53

recipes/ncurses/ncurses.inc
```diff
@@ -91,7 +91,7 @@ AUTO_PACKAGE_UTILS_RDEPENDS += "libncurses"
 FILES_${PN} = "${datadir}/tabset ${sysconfdir}/terminfo"
 RDEPENDS_${PN}-tools += "${AUTO_PACKAGE_UTILS_PROVIDES}"
 FILES_${PN}-terminfo = "${datadir}/terminfo ${libdir}/terminfo"
-FILES_${PN}-dev += "${bindir}/ncurses5-config"
+FILES_${PN}-dev += "${bindir}/ncurses?-config"
```
kimrhh: probably safer with [5-6] if you want to do pattern matching, to avoid random explodes when they make a ncurses7

recipes/gcc/gcc_5.3.0.oe
```diff
@@ -0,0 +1,15 @@
...
+#AUTO_PACKAGE_LIBS:>TARGET_CPU_x86_32 = " cilkrts vtv"
+#AUTO_PACKAGE_LIBS:>TARGET_CPU_i686 = " cilkrts vtv"
+
+FILES_${PN} += "${libdir}/libcc1.*"
+
+AUTO_PACKAGE_LIBS += "gomp-plugin-host"
+DEPENDS_${PN}-libgomp += "libgomp-plugin-host"
+RDEPENDS_${PN}-libgomp += "libgomp-plugin-host"
+
+#FILES_${PN}-
```
kimrhh: are we adding this in parallel with gcc-4.x ? since there is no deletion of the old version

kimrhh: kill!

recipes/ncurses/ncurses/ncurses-5.9-gcc-5.patch
```diff
@@ -0,0 +1,46 @@
...
++
+ $preprocessor $TMP 2>/dev/null \
+-| sed \
+-	-e 's/  / /g' \
+-	-e 's/^ //' \
+-	-e 's/_Bool/NCURSES_BOOL/g' \
++| sed -f $ED1 \
+ | $AWK -f $AW2 \
+ | sed -f $ED3 \
+ | sed \
```
kimrhh: why do we need a ncurses-5 patch for ncurses6 :P

sknsean:
target_cpu = d.get('TARGET_CPU');
...
if "cpu" == arg_name and (target_cpu == "arm" or target_cpu == "aarch64"):

PR #70, opened 2015-11-05
make: packageqa fixes

sknsean:

### DIFF
--- a/recipes/make/make.inc
+++ b/recipes/make/make.inc
@@ -12,6 +12,9 @@ RECIPE_TYPES = "native sdk"
 
 DEPENDS = ""
 
+DEPENDS_${PN} = "libc"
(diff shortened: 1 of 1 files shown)


recipes/make/make.inc
```diff
@@ -12,6 +12,9 @@ RECIPE_TYPES = "native sdk"
 
 DEPENDS = ""
 
+DEPENDS_${PN} = "libc"
+RDEPENDS_${PN}:sdk = "libc"
```
esben: Why the ':sdk' override?

sknsean:
Oh at first I did write RDEPENDS = "libc" and it complained, so i added the :sdk.
Later i added the _${PN} so i guess the :sdk is not needed then...

PR #71, opened 2015-11-05
glibc 2.22

sknsean:
Maybe for milestone 6.0.0


esben:
LGTM, but waiting for jenkins to give green light :)

esben:
You should probably apply the following patch to valgrind:
https://github.com/envytools/valgrind/commit/243de7f3e62e2d22fae1be08b5e28d5afef5f040

sknsean:
Do I need to include all the patches from https://sourceware.org/git/?p=glibc.git;a=shortlog;h=refs/heads/release/2.22/master in this PR?

esben:
Maybe.  But in that case, please don't convert the git commit log to a patch series.  Switch to git fetcher, and fetch the proper commit.  And keep it in separate commit, so we can play with revert on it in case of problems.

sknsean:
Glibc 2.23 have been released :-)
@esben do you want 2.22 or 2.23 for 6.0.0?

esben:
AFAIK, 2.23 is not released yet.  Until then, head of 2.22 release branch should be used.  Upgrading to 2.23 as soon as it is released.

esben:
LGTM

PR #72, opened 2015-11-10
kernel: clean up device tree handling

kimrhh:
The current device tree handling is a mess regarding naming. Try
to avoid confusion by consistently talking about "Device Tree Source"
when referring to .dts files, "Device Tree Binary" when referring to
.dtb files and "Device Tree Compiler" when referring to dtc.

It should now be able to handle a number of use-cases:

 - a compiled device tree is provided (.dtb)
 - a device tree source file should be compiled
	- using external dtc
	- using kernel dtc
	- using the in-kernel make target
 - flags can be given to the in-kernel dtc

Also clean up documentation, talking about kernel images when
explaing device tree variables.


classes/kernel.oeclass
```diff
@@ -15,10 +15,14 @@
 ##          DEPENDS_KERNEL_INITRAMFS to acquire the initramfs.
 ## @useflag ramdisk_image_compression Compression of the ramdisk image. Set to
 ##          none by default.
-## @useflag kernel_dtb Name of the compiled Device tree source (dts).
-## @useflag kernel_dtc Whether to use the linux device tree compiler (dtc).
+## @useflag kernel_dtb Name of the (already compiled) Device tree binary (dtb)
```
esben: I don't think it was ever meant as a way to provide an already compiled dtb file.  The idea was (AFAIR) to use USE_kernel_dtb to specify the dtb file that will be build by kbuild process.

classes/kernel.oeclass
```diff
@@ -15,10 +15,14 @@
...
+##          to be installed into boot directory. kernel_dts makes this
+##          unnecessary.
+## @useflag kernel_dtc Whether to use the device tree compiler (dtc) from the
+##          kernel source tree being built. If not set, dtc from oe-lite recipe
+##          will be used.
 ## @useflag kernel_dtc_flags Flags for the device tree compiler (dtc).
-## @useflag kernel_dtc_source Device tree source (dts).
+## @useflag kernel_dts Name of the Device tree source (dts) to be compiled
```
esben: Clearly a more appropriate name :)  But perhaps add "file", so it reads "Name of the Device tree source (dts) file to be compiled".

classes/kernel.oeclass
```diff
@@ -238,38 +239,47 @@ UIMAGE_KERNEL_OUTPUT = ""
...
 addhook kernel_devicetree_init to post_recipe_parse after set_useflags
 def kernel_devicetree_init(d):
-    kernel_dtc = d.get('USE_kernel_dtc')
     kernel_dtb = d.get('USE_kernel_dtb')
-    if kernel_dtc and kernel_dtc != 0:
-	kernel_dtc_source = d.get('USE_kernel_dtc_source')
-	dts = os.path.basename(kernel_dtc_source)
+    kernel_dtc = d.get('USE_kernel_dtc')
+    kernel_dts = d.get('USE_kernel_dts')
+    if (kernel_dtc and kernel_dts != 0) or kernel_dts != 0:
```
esben: Did you mean
```python
   if (kernel_dtc and kernel_dtc != 0) or kernel_dts != 0:
```
?

classes/kernel.oeclass
```diff
@@ -238,38 +239,47 @@ UIMAGE_KERNEL_OUTPUT = ""
...
+	kernel_dts = d.get('USE_kernel_dts')
+        dts_dir,dts = os.path.split(kernel_dts)
```
esben: minor whitespace mess...

classes/kernel.oeclass
```diff
@@ -238,38 +239,47 @@ UIMAGE_KERNEL_OUTPUT = ""
...
 	(dts_name, dts_ext) = os.path.splitext(dts)
 	if dts_ext != '.dts':
 	    dts_name = dts
-	d.set('KERNEL_DEVICETREE', dts_name + ".dtb")
+	d.set('KERNEL_DTB_FILE', dts_name + ".dtb")
+        if kernel_dts != d.get('KERNEL_DTB_DIR'):
```
esben: This seems like bad naming.  When would USE_kernel_dts (dts file name) be equal to KERNEL_DTB_DIR (Location of the compiled devicetree)?

Also, minor whitespace issue again.

classes/kernel.oeclass
kimrhh: ok so the make target that is automatically there when the corresponding .dts is present?

Think those two things should be seperated then, I do see a use for specifying a .dtb directly as well anyway

classes/kernel.oeclass
kimrhh: maybe, but that whole function seems to try to automatically figure out how things should be set up, I honestly don't like that. How about we just have flags for all settings, and then error out if the combinations are wrong? e.g. allow the users to specify DTBDIR with a USE as well

classes/kernel.oeclass
kimrhh: think this is due to the "try to figure out stuff" mentioned above, agree it doesn't look right, will take another look when you have replied :)

classes/kernel.oeclass
esben: That might be a good idea.  But the problem is coming up with USE flags / variables that both allow specifying all relevant configurations, while still making it possible to figure out how to do just that.  When that is achived, I agree with you, we should error out when something is wrong, instead of trying to figure out what to do.

kimrhh:
this version of the PR has been tested with various setups, including external dtb file, no dtb (x86) and various settings for in-kernel dts/dtb. I think it is as ready as it gets for merging :)

esben:
Yes, let's do it like this :+1:

PR #73, opened 2015-11-17
patchelf: Added

moths:
Patchelf for patching elf files. Used in sdk setup script for setting ld.so path.

### DIFF
--- a/recipes/patchelf/patchelf.inc
+++ b/recipes/patchelf/patchelf.inc
@@ -1,0 +1,12 @@
+SRC_URI = "http://nixos.org/releases/${PN}/${P}/${P}.tar.bz2"
+LICENSE = "GPLv3"
+SUMMARY = "Tool to allow editing of RPATH and interpreter fields in ELF binaries"
+
+inherit autotools c++ c
+
+RECIPE_TYPES = "native sdk"
+
+DEP = "host:libc host:libgcc host:libm host:libstdc++"
+RDEPENDS_${PN} += "${DEP}"
+DEPENDS_${PN} += "${DEP}"
+EXTRA_OECONF += "CFLAGS=-static CXXFLAGS=-static"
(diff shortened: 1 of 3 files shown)


recipes/patchelf/patchelf.inc
```diff
@@ -0,0 +1,12 @@
+SRC_URI = "http://nixos.org/releases/${PN}/${P}/${P}.tar.bz2"
+LICENSE = "GPLv3"
+SUMMARY = "Tool to allow editing of RPATH and interpreter fields in ELF binaries"
+
+inherit autotools c++ c
+
+RECIPE_TYPES = "native canadian-cross"
```
esben: Is it really canadian-cross, or is it just sdk?  Is it the canadian-cross patchelf built able to modify any ELF binaries, or is it restricted to modifying MACHINE architecture ELF binaries?

recipes/patchelf/patchelf.inc
```diff
@@ -0,0 +1,12 @@
...
+
+DEP = "host:libc host:libgcc host:libm host:libstdc++"
+RDEPENDS_${PN} += "${DEP}"
+DEPENDS_${PN} += "${DEP}"
+EXTRA_OECONF += "CFLAGS=-static CXXFLAGS=-static"
```
esben: Any special reason for static linking here?

recipes/patchelf/patchelf.inc
esben: I took a brief look at it, and I believe that any patchelf command can be used on any ELF binary.  So it should be changed to sdk instead of canadian-cross.

moths: Yes you are correct :)

recipes/patchelf/patchelf.inc
moths: Using it to patch the sdk binaries in an sdk-install script. Having it depend on the sdk provided libc/stdc++ will require a wrapper script for patchelf only. The benefit would be saving <1 MB (32 bit arm). But ads complexity I do not want.

esben: Fair enough :+1:

PR #76, opened 2015-12-01
image_passwd.oeclass: support use flag creation of users/groups

hundeboll:
Add post rstage functions to create users and groups listed in
USE_image_passwd_users and USE_image_passwd_groups, respectively.

In order to actually include the created users/groups in the final
image, the functions are run as postfuncs in the rstage step, and they
work on etc/passwd and etc/group in the rstage dir. This is needed
because the do_compile step moves the contents from ./rstage to ./image
and creates the image without possibility to inject python functions in
between. This also requires changing the existing
image_preprocess_passwd() function to run as rstage postfunc instead of
"IMAGE_PREPROCESS_FUNCS", in order to have its created users/groups
available when parsing the mentioned use flags.

Example user configartion in e.g. a distro configuration:
DISTRO_USE_image_passwd_users = \
 "user1:$1$L9mzJfTZ$2ED5as2K2yZ98CN/BQuy1.:/home/user1:/bin/sh:audio"
DISTRO_USE_image_passwd_users += \
 "nobody:$1$L9mzJfTZ$2ED5as2K2yZ98CN/BQuy1.::/bin/nologin:"

Example group configuration in e.g. a distro configuration:
DISTRO_USE_image_passwd_groups = "nobody,spi,i2c"

### DIFF
--- a/classes/image_passwd.oeclass
+++ b/classes/image_passwd.oeclass
@@ -4,10 +4,10 @@
 ## /etc/passwd provided by one or more recipes.
 
 require conf/passwd.conf
+require conf/makedevs.conf
 
-IMAGE_PREPROCESS_FUNCS += "image_preprocess_passwd"
-
-image_preprocess_passwd () {
+do_rstage[postfuncs] += "do_rstage_passwd_concat"
+do_rstage_passwd_concat () {
 	cwd=`pwd`
 	if [ -d $cwd/${passwddir} ]; then
 		for f in $cwd/${passwddir}/* ; do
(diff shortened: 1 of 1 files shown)


esben:
The USE flag format does not support specifying UID,GID.  I believe it could be handy in some cases.  And it would definitely be nice to not have to change the USE flag syntax if/when adding that.

Why not just use the passwd(5) and group(5) syntax instead (newline separated)?
An empty UID,GID could be allowed and handled as you do now.

hundeboll:
Sure, that would also simplify the implementation a bit

PR #77, opened 2015-12-01
image_mdev.oeclass: support use flag entries in mdev.conf

hundeboll:
Add post rstage function to add entries from USE_image_mdev_nodes
in etc/mdev.conf.

To make sure the entries are added after any recipe defined sections,
the existing image_preprocess_mdev() function in IMAGE_PREPROCESS_FUNCS
is changed to run as post rstage, as the newly created do_rstage_mdev()
function is written in python.

The function reads entries in USE_image_mdev_nodes and creates appends
lines to etc/mdev.conf with the configured match-strings and group-spec,
bases on the following syntax:
<match>:<group>

Example setting in e.g. distro configuration:
DISTRO_USE_image_mdev_nodes = "spidev.*:spi"

### DIFF
--- a/classes/image_mdev.oeclass
+++ b/classes/image_mdev.oeclass
@@ -23,6 +23,18 @@ image_preprocess_mdev () {
 	rm -rf $cwd/${mdevdir}
 }
 
+# USE_image_mdev_lines: lines to add to /etc/mdev.conf
+# syntax:  <match> <user/uid>:<group/gid> <mode> [=<subfolder>]
+# example: spidev.* 0:spi 660
+CLASS_FLAGS += "image_mdev_lines"
+IMAGE_PREPROCESS_FUNCS:>USE_image_mdev_lines = " image_preprocess_mdev_lines"
+image_preprocess_mdev_lines() {
+    cwd=`pwd`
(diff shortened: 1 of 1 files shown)


classes/image_mdev.oeclass
```diff
@@ -23,6 +24,33 @@ image_preprocess_mdev () {
...
+
+    mdev_file = open(mdev_path, 'a')
+    mdev_nodes = d.get("USE_image_mdev_nodes") or ""
+
+    if mdev_nodes:
+        mdev_file.write("\n# USE_image_mdev_nodes\n")
+
+    for mdev_node in mdev_nodes.split():
+        match,group = mdev_node.split(':')
+        mdev_line = "{:16s}0:{:8s} 660\n".format(match, group)
```
esben: Wouldn't it make sense to be able to control user and mode in the USE flag also?

hundeboll: Yeah, I can extend the syntax with three more fields:
<match>:<user/uid>[:<group/gid>[:<mode>[:subfolder]]]
?

esben: That should do it.

hundeboll:
Okay, i changed the commit to simply echo the contents of the use flags into /etc/mdev.conf. There was no real need for the python functionality in this case.

PR #78, opened 2015-12-01
image_depmod: add class to run depmod on image modules

hundeboll:
When collecting modules from out-of-tree builds, one must run depmod at
image creation time to resolve module dependencies. E.g. with wifi
drivers from linux-backports fails to load at first boot if dependencies
are missing.


esben:
RAID1?  I think we might take the chance of just having this functionality in one class though :smile:

hundeboll:
Pushed without backup class :)

PR #82, opened 2016-01-05
Log and tmp file handling

moths:
lib/oelite: Use date-time stamp for .log and .run files
    
    Example:
    new: do_configure.20151213141516.run   (2015/12/13 14:15:16)
    old: do_configure.32768.run
    
    This replaces the PID based filename which was not uniqe and
    resulted in .log and .run files being overwritten (especially if running
    oe bake in docker). The UTC timestamp is more likely to generate a uniqe
    file. (except on build hosts with broken/no RTC, during local hard time
    adjustments, leap second, ...)

### DIFF
--- a/lib/oelite/function.py
+++ b/lib/oelite/function.py
@@ -133,7 +133,7 @@ class ShellFunction(OEliteFunction):
         runfile.write("#!/bin/bash -e\n\n")
         if os.path.exists(runsymlink) or os.path.islink(runsymlink):
             os.remove(runsymlink)
-        os.symlink(runfn, runsymlink)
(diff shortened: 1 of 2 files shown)


lib/oelite/function.py
```diff
@@ -133,7 +133,7 @@ def __call__(self):
         runfile.write("#!/bin/bash -e\n\n")
         if os.path.exists(runsymlink) or os.path.islink(runsymlink):
             os.remove(runsymlink)
-        os.symlink(runfn, runsymlink)
+        os.symlink(os.path.basename(runfn), runsymlink)
```
esben: Small thing, but nice :)

PR #83, opened 2016-01-05
Remove zip and libz from provides.conf

moths:
Remove from provide.conf and therby use the oe-lite build zip and libz versions.

zip and libz is not directly mentioned in the setup/* files so no cleanup needed there.

### DIFF
--- a/conf/provided/all.conf
+++ b/conf/provided/all.conf
@@ -19,6 +19,7 @@ native:gzip native:bzip2 native:tar native:unzip native:xz \
 native:patch native:quilt native:diffstat \
 native:perl native:perl-runtime \
 native:python-runtime \
+native:ruby \
 native:texinfo native:doxygen \
 native:texlive-extra-utils native:texlive-latex-extra native:latex-xcolor \
 native:util-linux native:coreutils \
--- a/conf/provided/all.conf
+++ b/conf/provided/all.conf
@@ -15,7 +15,7 @@ native:m4 \
 native:git native:cvs native:svn native:mercurial native:bzr \
 native:autoconf native:automake native:gnu-config \
 native:bc \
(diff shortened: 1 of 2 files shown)


conf/provided/all.conf
```diff
@@ -15,10 +15,11 @@ native:m4 \
 native:git native:cvs native:svn native:mercurial native:bzr \
 native:autoconf native:automake native:gnu-config \
 native:bc \
-native:gzip native:bzip2 native:tar native:unzip native:zip native:xz \
+native:gzip native:bzip2 native:tar native:unzip native:xz \
 native:patch native:quilt native:diffstat \
 native:perl native:perl-runtime \
 native:python-runtime \
+native:ruby \
```
esben: This is not mentioned in commit message.  Is it added on purpose?

conf/provided/all.conf
```diff
@@ -28,4 +29,5 @@ native:mtd-utils-mkfs-jffs2 \
 native:wget \
 native:gawk native:sed native:grep \
 native:libiconv \
+native:libacl1-dev \
```
esben: This is not mentioned in commit message.  Is it added on purpose?

conf/provided/all.conf
kimrhh: "conf/provided/all: Add ruby, build time dependency for qtwebkit" :)

esben: Hehe, didn't see that there were more than 1 commit.

conf/provided/all.conf
moths: ok removed, it is relate to native:fakeroot. Without it fakeroot ends up building a filesystem with wrong permissions. Causing fx. /usr to only work for root, not normal users. But it belongs in another patch

PR #85, opened 2016-01-10
aarch64 support / updating of config.guess .sub

sknsean:


recipes/gcc/gcc-5.3/aarch64-multilib-dirnames.patch
```diff
@@ -0,0 +1,12 @@
+--- gcc-5.3.0/gcc/config/aarch64/t-aarch64-linux.bak	2016-01-07 13:52:31.113702383 +0000
++++ gcc-5.3.0/gcc/config/aarch64/t-aarch64-linux	2016-01-07 13:55:39.470923680 +0000
+@@ -22,7 +22,7 @@
+ LIB1ASMFUNCS = _aarch64_sync_cache_range
+ 
+ AARCH_BE = $(if $(findstring TARGET_BIG_ENDIAN_DEFAULT=1, $(tm_defines)),_be)
+-MULTILIB_OSDIRNAMES = mabi.lp64=../lib64$(call if_multiarch,:aarch64$(AARCH_BE)-linux-gnu)
++MULTILIB_OSDIRNAMES = mabi.lp64=.$(call if_multiarch,:aarch64$(AARCH_BE)-linux-gnu)
```
esben: Are you sure this works?  Won't this place libraries in bindir?

recipes/glibc/glibc.inc
```diff
@@ -43,6 +43,7 @@ do_configure_siteinfo() {
 }
 
 EXTRA_OECONF += "\
+	libc_cv_slibdir='/lib' \
```
esben: Hmm...  This looks very much like the multilib fixing we still have not agreed on...

classes/autotools.oeclass
```diff
@@ -48,7 +48,13 @@ autotools_configure[emit] = "do_configure"
 do_configure () {
     do_configure_autotools
 }
+AUTOCONFDIR = ""
+UPDATE_GNUCONF = ""
+UPDATE_GNUCONF:TARGET_CPU_aarch64 = "rm -f ${S}${AUTOCONFDIR}/config.{guess,sub}; cp ${TOPDIR}/meta/core/scripts/config.{guess,sub} ${S}${AUTOCONFDIR}"
```
esben: Please move this "one-liner" into do_configure_autotools, and use UPDATE_GNUCONF as a 0/1 flag instead.  This makes it easier to update config.{guess,sub} in other situations.

recipes/glibc/glibc.inc
sknsean: Without it libs will be installed in /lib64

esben: Yes, so you say :)  But I am still waiting for instructions on how to reproduce this.  It seems strange that such a fundamental problem has been unseen for such a long time.

I still suspect that it is caused by the output from "$CC -print-multi-os-directory" of the native compiler on specific Linux distros.  Maybe we need to ignore -print-multi-os-directory for native compilers, but I don't think it should be ignored for cross and sdk-cross compilers that we build ourselves.

esben:
If you could split out the gcc and glibc changes into a new PR, I could merge the other commits.  You just need to update their commit message with something a bit more descriptive than "Added Aarch64 support"

sknsean:
gcc splitted to #104 
glibc changes moved to sdkfix #64

recipes/glibc/glibc.inc
sknsean: Moved to #64

classes/autotools.oeclass
sknsean: What do you mean by "move this "one-liner" into do_configure_autotools" it's used inside the do_configure_autotools?

sknsean: Like this?

classes/autotools.oeclass
```diff
@@ -48,7 +48,16 @@ autotools_configure[emit] = "do_configure"
 do_configure () {
     do_configure_autotools
 }
+AUTOCONFDIR = ""
+UPDATE_GNUCONF ?= "0"
+UPDATE_GNUCONF:TARGET_CPU_aarch64 = "1"
+
 do_configure_autotools () {
+    if [ -a "${UPDATE_GNUCONF}" != "0" ] ; then
```
esben: I think you have an "-a" to much in there.  But otherwise, yes, this is better :)

classes/autotools.oeclass
```diff
@@ -48,7 +48,16 @@ autotools_configure[emit] = "do_configure"
...
+        rm -f ${S}${AUTOCONFDIR}/config.{guess,sub}
+        cp ${TOPDIR}/meta/core/scripts/config.{guess,sub} ${S}${AUTOCONFDIR}
```
esben: Oh, there is a problem with this approach.  By taking files out from manifest like this, you are bypassing the signature system.  So changes to config.{guess,sub} is not tracked at all, which is clearly not a good thing.

To fix this, we probably need to create a recipe for the config.{guess,sub} files, and then add a dependency to this in CLASS_DEPENDS if UPDATE_GNUCONF is set, and then copy from stage instead.

sknsean: less rebuilding :-)

esben: Well, we tend to favour build correctness :wink:

sknsean:
Fun times :-)
A package like autoconf have 2 config.guess and 2 config.sub files...

sknsean:
@esben ping

classes/autotools.oeclass
```diff
@@ -48,7 +48,22 @@ autotools_configure[emit] = "do_configure"
 do_configure () {
     do_configure_autotools
 }
+
+AUTOCONFDIR = ""
+S_AUTOCONFDIR = ""
```
esben: You probably need to add a comment on the meaning of these two variables, so there is a chance that people will figure out how to use this.

sknsean: A package like autoconf have 2x config.{guess,sub} files.
This adds support for overwriting config.* in two seperate dirs...
Maybe one day a package have 3 locations for config.* files. Maybe this could be done differently... Any ideas?

esben: Like a list of dirs?  The current approach is definitely not the best idea.

sknsean: some example pls :-)

esben:
```sh
for autoconfdir in ${AUTOCONFDIRS} ; do
    rm -f ${S}$autoconfdir/config.{guess,sub}
    ....
done
```

PR #89, opened 2016-01-17
Cherry-pick of commits from old PR #13

esben:
This is a cherry pick of most of the commits in the old PR #13 from @diorcety.

### DIFF
--- a/recipes/binutils/binutils.inc
+++ b/recipes/binutils/binutils.inc
@@ -133,9 +133,6 @@ RDEPENDS_${PN}-strip = "libbfd"
 RDEPENDS_${PN}-ar = "libbfd"
 RDEPENDS_${PN}-strings = "libbfd"
 
-# The binutils package is used to pull in all provided util features
-RDEPENDS_${PN} = "${AUTO_PACKAGE_UTILS_PROVIDES}"
-
 # We need to set a different priority than the highest gcc recipe priority, so
 # that we don't get into multiple provider problems with libbfd.
 DEFAULT_PREFERENCE = "2"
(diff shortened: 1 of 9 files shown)


recipes/binutils/binutils.inc
```diff
@@ -12,12 +12,12 @@ BUGTRACKER = "http://sourceware.org/bugzilla/"
...
 RECIPE_TYPES = "cross sdk-cross machine canadian-cross"
 
-inherit autotools make-vpath gettext
+inherit autotools make-vpath
 
 require conf/fetch/gnu.conf
 SRC_URI = "${GNU_MIRROR}/binutils/binutils-${PV}.tar.bz2 "
 
-DEPENDS = "native:flex native:bison"
+DEPENDS = "native:flex native:bison native:makeinfo native:m4"
```
esben: @diorcety where do you expect to get native:makeinfo from?  Did you send a PR with makeinfo recipe, or should it be added to ASSUME_PROVIDED?

diorcety: https://github.com/diorcety/oe-lite-base/commit/b2f8aefabd3f3dfa29b5c876afc9b59f15770fcc from https://github.com/oe-lite/base/pull/22

esben: Ok.  We should get that merged then :)

But what about adding native:makeinfo to conf/provided/minimal.conf also?

diorcety: Can be. But all these PRs have goal to have the minimal provided tool (a minimal minimal)

esben: Yes.  Maybe we should only add it to default.conf and all.conf instead then.
That is probably both the right thing, and will allow us to build test this PR without the texinfo recipe merged in OE-lite/base.

diorcety: I agreed. So we let the PR like this and i create the PR for texinfo into base?

esben: I think you should add native:makeinfo to default.conf and all.conf in this PR, so we can merge it independently of changes in base.

And creating a PR for texinfo in base would be great.

PR #90, opened 2016-01-19
Glibc 2.21 upstream stable branch fixes and gcc 4.9.3

moths:


kimrhh:
wouldn't it be better to upgrade to glibc-2.22 while we are at it? instead of adding all the patches :)

esben:
No, not for 5.0.  GCC and glibc upgrade is 6.0 stuff.

moths:
Staying on 2.21 for 5.0.0 release seems the right thing to do. The patches are not 2.21 to 2.22, but 2.21 stable branch fixes.

kimrhh:
@esben eh, this PR is upgrading GCC ;) but if you consider 2.22 a more unstable upgrade then fine, sounds like it is based on mortens input

esben:
gcc 4.9.3 is a bugfix release, glibc 2.22 is a normal release, so I agree with @moths

kimrhh:
try synchronizing this, now that kexec-tools should be fixed (hat tip @sknsean )

esben:
I'll merge it now, but I think we should drop gcc 4.9.2 before release of OE-lite/core 5.0.0.

PR #100, opened 2016-02-17
libtool: Split native build

diorcety:
Add patches for native & cross for working correctly with oe-lite


esben:
@diorcety could you try and rebase this to current master, and see if it can pass our buildbot?

PR #101, opened 2016-02-18
Glibc221 fixes

moths:
Switch to git and update to include the fix for remote getaddrinfo() stack-based buffer overflow. with switch to git from #71


sknsean:
I believe you can delete the glibc_2.21.oe.sig otherwise ok :-)

esben:
Yes, the .sig file can be dropped, but otherwise (pending green light) this is ready to be merged for 5.0.0 :)

PR #102, opened 2016-02-22
gcc: Add support for machine builds

esben:


kimrhh:
some more work on this, I don't have time to look at the remaining bits. There is a bit of package-qa and still some target prefixed binaries (but also non-prefixed now)

[0002-gcc-avoid-sysroot-for-machine-build.txt](https://github.com/oe-lite/core/files/146717/0002-gcc-avoid-sysroot-for-machine-build.txt)
[0001-gcc-avoid-target-prefixed-binaries-for-machine.txt](https://github.com/oe-lite/core/files/146718/0001-gcc-avoid-target-prefixed-binaries-for-machine.txt)

esben:
Should we postpone this to after 6.0 ?

kimrhh-prevas:
yep, it does not have any strong priority in our end currently :)

PR #103, opened 2016-02-22
gcc: Remove 4.9.3 leftovers

esben:


esben:
I am fast-track merging this, as we cannot even parse recipes without it.

PR #104, opened 2016-02-23
gcc: Added Aarch64 support

sknsean:
Signed-off-by: Sean Nyekjaer <sean@nyekjaer.dk>


recipes/gcc/gcc-5.3/aarch64-multilib-dirnames.patch
```diff
@@ -0,0 +1,12 @@
+--- gcc-5.3.0/gcc/config/aarch64/t-aarch64-linux.bak	2016-01-07 13:52:31.113702383 +0000
++++ gcc-5.3.0/gcc/config/aarch64/t-aarch64-linux	2016-01-07 13:55:39.470923680 +0000
+@@ -22,7 +22,7 @@
+ LIB1ASMFUNCS = _aarch64_sync_cache_range
+ 
+ AARCH_BE = $(if $(findstring TARGET_BIG_ENDIAN_DEFAULT=1, $(tm_defines)),_be)
+-MULTILIB_OSDIRNAMES = mabi.lp64=../lib64$(call if_multiarch,:aarch64$(AARCH_BE)-linux-gnu)
++MULTILIB_OSDIRNAMES = mabi.lp64=.$(call if_multiarch,:aarch64$(AARCH_BE)-linux-gnu)
```
sknsean: I honestly can remember why it ended up like this :-) But I can see your point so I will do some testing.

sknsean:
I guess this is ready :-)

skny-prevas:
@esben what do you think of the proposal?

esben:
Looking better.

A final remark though: In b6ee663010ea663e4def99d1d8b59ea0e33d7f99 you state that "GCC will no longer provide -print-multi-os-directory".  But that is only true for OE-lite built GCC.  We are still using native:gcc from host OS, so we should be on the look out for problems related to that.

But let's try this approach.

PR #113, opened 2016-03-09
binconfig.oeclass: fix file globbing in do_install_binconfig_fixup

hundeboll:
When running glob.glob() in python, the call does a search on the file
system to expand wildcards to matching files. This failed in binconfig
due to globing relative to the source directory and the removal of
heading slashes (i.e. '/usr/bin/*-config' -> 'usr/bin/*-config'.

Fix this by changing working directory prior to the globing, so that the
search is done relative to the install directory.


esben:
Oops, good one.  :+1:

PR #114, opened 2016-03-10
binconfig_stage.oeclass: catch errors on symlinking and handle them.

hundeboll:
we sometimes get a throw on symlinking. We usually have to rebuild almost
every thing. Instead this commit tries simply to force symlinking it.


esben:
I don't think this is such a good idea.  We do not guarantee a deterministic ordering of staging of package.  And even if we did, I don't think it would be in a way that would obvious to developers.

I don't remember seeing such problems at all.  Do you have some kind of idea what is causing it?

PR #116, opened 2016-03-10
fakeroot: add correct dependencies.

hundeboll:
fakeroot depends on host libacl1-dev to set correct permissions on
files/folders.

If libacl1-dev is not available, all folders/files are root-only
writable.

### DIFF
--- a/recipes/fakeroot/fakeroot.inc
+++ b/recipes/fakeroot/fakeroot.inc
@@ -11,6 +11,7 @@ HOMEPAGE = "http://fakeroot.alioth.debian.org"
 inherit autotools
 
 RECIPE_TYPES="native"
+DEPENDS = "native:libacl1-dev"
 
 require conf/fetch/debian.conf
 SRC_URI = "${DEBIAN_MIRROR}/main/f/fakeroot/fakeroot_${PV}.orig.tar.bz2"


classes/fakeroot.oeclass
```diff
@@ -3,7 +3,7 @@
 ## By inheriting this class, it is ensures that the native version of fakeroot
 ## build, is used instead of fakeroot from the build system.
 
-CLASS_DEPENDS += "native:fakeroot"
+CLASS_DEPENDS += "native:fakeroot native:libacl1-dev"
```
esben: Uh, why is that libacl1-dev and not just libacl1?  We really should normally not depend on lib*-dev in DEPENDS.  All lib* packages is responsible for pulling in their own lib*-dev.  If they don't do that, they are broken.

kimrhh: we don't have a recipe for it, it is a build-host dep, and without it fakeroot messes up at least for debian like distroes.

kimrhh: doesn't look like something that would be hard to replace with a proper recipe though :)

esben: Is this change targeted at native:fakeroot from host distro?  Or native:fakeroot from OE-lite/core recipe?

hundeboll: You are right, I have moved the dependency from the class to the recipe.

Building fakeroot with libacl1-dev *not* installed:
```
checking sys/acl.h usability... no
checking sys/acl.h presence... no
checking for sys/acl.h... no
```

Building fakeroot with libacl1-dev installed:
```
checking sys/acl.h usability... yes
checking sys/acl.h presence... yes
checking for sys/acl.h... yes
```

esben:
It still hurts my eyes to see a -dev depedency in DEPENDS.  It needs to be fixed when we get our own libacl recipe.

PR #118, opened 2016-03-10
class/kernel: refactor use of kernel_dtc and kernel_dtb

hundeboll:
Add placeholder variables for compile/install/deploy postfuncs to make
conditional calling of such functions easier.

Facktor out some of the functions from do_install() and do_deploy() to
make those two a little shorter.


kimrhh:
this should die in favor of #72

PR #120, opened 2016-03-10
conf/u-boot.conf: make UBOOT_IMAGE use flag dependent.

hundeboll:
Use use flags instead of a variable override in machine.conf or whatever.
Variables in machine.conf will propagate to all signatures whereas use
flags is only relavant for specific recipes.


conf/u-boot.conf
```diff
@@ -3,17 +3,19 @@
...
 
 DEFAULT_USE_uboot_config ?= "${MACHINE}_config"
 DEFAULT_USE_uboot_extra_env ?= ""
+DEFAULT_USE_uboot_image ?= "u-boot.bin"
 
 COMPATIBLE_MACHINES = ".*"
 
 ## @var UBOOT_IMAGE Name of the image to build. Is u-boot.bin by default.
-UBOOT_IMAGE ?= "u-boot.bin"
+UBOOT_IMAGE ?= "${USE_uboot_image}"
```
esben: Could you add
```
DEFAULT_USE_uboot_image = "u-boot.bin"
```
Which should be a sane default, and give us backwards compatibility.

hundeboll: See line 13 above...

hundeboll:
I am wondering if I should add a use flag for `UBOOT_IMAGE_FILENAME` (and possibly `UBOOT_IMAGE_BASE`) as those are likely to change if `USE_uboot_image` is configured to something else than the default...

conf/u-boot.conf
esben: Bah :)

esben:
Sounds like a good idea.

hundeboll:
Updated with more use flags

PR #123, opened 2016-03-10
lib/oelite/fetch/url: traverse mirror list while signature does not match

hundeboll:


esben:
Yes, it seems to be in place already :)

PR #127, opened 2016-03-10
setup: updates to ubuntu setup files

hundeboll:


esben:
It is not fair to say that it makes it possible to auto install oe-lite deps.  It will always do that.  But I don't think it is risky, so let's just do it that way.

PR #129, opened 2016-03-11
binutils: make libbfd depend on libdl

hundeboll:
With recent glibc, libbfd depends on libdl. Without this, oprofile fails
do_configure with

arm-cortexa9neon-linux-gnueabi/bin/ld: warning: libdl.so.2, needed by arm-cortexa9neon-linux-gnueabi/sysroot/usr/lib/libbfd.so,
not found (try using -rpath or -rpath-link)


esben:
Did packageqa show this problem?  And did it verify the fix?

hundeboll:
packageqa shows
```
Package binutils-libbfd NEEDED: libz.so.1 libc.so.6 libdl.so.2
```

PR #130, opened 2016-03-24
Memory use reduction

Villemoes:
oe-lite is a pig with its memory consumption. I've played around with dumping some internal state to see where all that memory is going and then tried to do something about it. This isn't a real pull request (and I'm going to close it immediately) since it contains some of the debugging stuff I've used so that other people can play around with it.

The TL;DR is that in the simple case I've been using (oe bake gnutls), the memory use after task metadata hash computation drops from around 350 MiB to around 192 MiB. I fully expect to see a proportional saving when doing a larger task, but haven't gotten around to testing it yet.


esben:
Good work!  I look forward to a new PR which we can actually merge :)

What impact does this have on execution time?

PR #132, opened 2016-03-29
Faster metadata hashing

Villemoes:
We spend longer on the initial part of 'oe bake' than we should; that's particularly annoying when there's actually nothing to do. These are just a few low-hanging fruits, and I freely admit that the last patch is really ugly, but OTOH that's what gives the largest improvement.

Master + the just the first patch:

$ time oe bake gnutls
Adding recipes to cookbook: 233 / 233 [100 %]
Building dependency tree time 1.114 seconds
Calculating task metadata hashes: 405 / 405 [100 %]
Calculation task metadata hashes time 6.266 seconds
Nothing to do

real    0m9.760s
user    0m9.500s
sys     0m0.244s

With all these:

$ time oe bake gnutls
Adding recipes to cookbook: 233 / 233 [100 %]
Building dependency tree time 1.117 seconds
Calculating task metadata hashes: 405 / 405 [100 %]
Calculation task metadata hashes time 3.984 seconds
Nothing to do

real    0m7.393s
user    0m7.180s
sys     0m0.196s

We're still spending more than half the time on the hash computations, so there's probably more to get from there.

### DIFF
--- a/lib/oelite/baker.py
+++ b/lib/oelite/baker.py
@@ -309,8 +309,7 @@ class OEliteBaker:
                 die("No such task: %s: %s"%(thing, e.__str__()))
             except oebakery.FatalError, e:
                 die("Failed to add %s:%s to runqueue"%(thing, task))
-        if self.debug:
-            timing_info("Building dependency tree", start)
+        timing_info("Building dependency tree", start)
 
         # Generate recipe dependency graph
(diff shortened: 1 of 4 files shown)


esben:
Last patch is fine.  The entire code is not really so pretty, so it just adapts to the style ;)

PR #133, opened 2016-04-18
Memory use reduction, v4

Villemoes:
This is morally identical to the patches in PR #131, but rebased on top of current master (to fix up some conflicts with PR #132 which was merged in the meantime). Submitting again hoping that the openssl issue is now gone. Original pull request text:

These patches reduce the memory footprint of oelite by about 35%, measured immediately after the task metadata hash computation is done. They do not have any measurable effect on the run time of a "oe bake machine" when machine is already fully built (see below for details). I do believe that the runtime of an actual build will decrease when the kernel has more RAM (on my machine, on the order of an extra 1.5% of total memory, or about 240MiB) to play with for page cache etc. instead of having it tied up in a process which will mostly never use it again.

The first three patches are not directly related to this goal and could/should go in on their own, but the first (get str objects from sqlite, not unicode) is a prerequisite for some of the following patches, so I included them here.

I cherry-picked these patches to a random customer project and did a full 'rm -rf tmp/ && oe bake machine' to check that I didn't break anything, but obviously I cannot be sure that that would exercise all code paths, so there may certainly be stuff to fix - I do believe that the possible memory savings are worth it, despite whatever pain we may encounter.

So, with machine fully built and everything nicely cached, I watched the memory consumption in top. Before, it peaked at 4.5%, after at 3.0%. This is consistent with the numbers I got from "within" python, using the meliae module to inspect the memory use of all live objects. Timings are

$ time oe bake machine
Adding recipes to cookbook: 301 / 301 [100 %]
Calculating task metadata hashes: 1244 / 1244 [100 %]
Nothing to do

real 0m35.678s
user 0m35.236s
sys 0m0.456s

$ time oe bake machine
Adding recipes to cookbook: 301 / 301 [100 %]
Calculating task metadata hashes: 1244 / 1244 [100 %]
Nothing to do

real 0m35.638s
user 0m35.308s
sys 0m0.344s

### DIFF
--- a/lib/oelite/cookbook.py
+++ b/lib/oelite/cookbook.py
@@ -32,6 +32,7 @@ class CookBook(Mapping):
         self.db = sqlite.connect(":memory:")
         if not self.db:
             raise Exception("could not create in-memory sqlite db")
+        self.db.text_factory = str
         self.dbc = self.db.cursor()
         self.init_db()
         self.recipes = {}
--- a/lib/oelite/meta/dict.py
+++ b/lib/oelite/meta/dict.py
@@ -94,7 +94,7 @@ class DictMeta(MetaData):
 
     def trim_expand_cache(self, var):
         for (cached_var, (cached_val, deps)) in self.expand_cache.items():
-            if cached_var == var or var in deps:
+            if cached_var == var or (deps and var in deps):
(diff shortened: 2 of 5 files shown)


esben:
Great work, thanks!

PR #134, opened 2016-04-27
oelite: runq.py: add an index to the recdepend table

Villemoes:
During my "make it faster" efforts, I stumbled on this, which I'm sending by itself since I believe it is highly unlikely to break anything and obviously a huge win.

Creating an index cuts over 75% of the "Building dependency tree
time", from ~6m04s to ~1m29s.


esben:
Awesome!

PR #135, opened 2016-04-27
Make it even faster

Villemoes:
Here's a bunch of semi-low hanging fruit. I think they are all "obviously" correct, but someone should probably check. For me, an "oe bake world" on current master takes ~4m54s to get to the "Do you want to continue" prompt, with these patches, that's down to ~2m10s. The biggest contributor is  	0e1f20a, but the others also cut some seconds here and there.

### DIFF
--- a/lib/oelite/parse/oeparse.py
+++ b/lib/oelite/parse/oeparse.py
@@ -582,13 +582,13 @@ class OEParser(object):
             return
         if not os.path.isabs(filename) and not filename.endswith(".oeclass"):
             filename = os.path.join("classes", "%s.oeclass"%(filename))
-        if not "__inherits" in self.meta:
-            self.meta["__inherits"] = [filename]
-        else:
-            __inherits = self.meta["__inherits"]
+        __inherits = self.meta.get("__inherits")
+        if __inherits:
             if filename in __inherits:
                 return
-            self.meta["__inherits"].append(filename)
+            __inherits.append(filename)
+        else:
+            self.meta["__inherits"] = [filename]
         self.include(filename, p, require=True)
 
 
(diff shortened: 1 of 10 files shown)


lib/oelite/parse/oeparse.py
```diff
@@ -582,13 +582,13 @@ def inherit(self, filename, p):
             return
         if not os.path.isabs(filename) and not filename.endswith(".oeclass"):
             filename = os.path.join("classes", "%s.oeclass"%(filename))
-        if not "__inherits" in self.meta:
-            self.meta["__inherits"] = [filename]
-        else:
-            __inherits = self.meta["__inherits"]
+        __inherits = self.meta.get("__inherits")
+        if __inherits:
```
esben: What about
```python
__inherits = self.meta.setdefault("__inherits", [])
if filename not in __inherits:
    __inherits.append(filename)
```
which also takes care of the else clause below.  The `[filename]` is processed in all cases though, causing an overhead...

esben:
More awesomeness, thanks!

I guess you sort of proved with 0e1f20a that the design choice to use sqlite was a mistake.
The idea was to get more clear and maintainable code with better performance.  It would have been better to have stayed with doing it by hand in Python.

I am ready to merge it, but could you take a quick look at my suggestion for fd6f939ddcf8dfa551ce4b60509a31a140acc5b8 ?

lib/oelite/parse/oeparse.py
Villemoes:
```
ravi@prevas-ravi:/mnt/xfs/devel/oe-lite/meta/core$ git grep setdefault
ravi@prevas-ravi:/mnt/xfs/devel/oe-lite/meta/core$ 
```


I suppose I know what you mean, but I don't think DictMeta has a method for 'get the value if it exists, otherwise set it to this and return that'... but I don't really care either way, so feel free to omit this patch and/or replace with something else

esben:
I won't hold this back anymore :)

PR #139, opened 2016-06-03
classes/binconfig-stage: create destination dirs before symlinking

hundeboll:
When staging packages that inherit binconfig, binconfig_stage_fixup() is
called to create a symlink:
./stage/cross/${bindir}/${PN}-config -> ./stage/machine/${bindir}/${PN}-config

When binconfig_stage_fixup() is called the destination directory for the
symlink (./stage/cross/${bindir}) might not exist, in which case the
symlink creating fails. This is the case when the first dependency being
stages uses binconfig-fixup.

Simply fix this by always creating the destination directory before
symlinking.

### DIFF
--- a/classes/binconfig-stage.oeclass
+++ b/classes/binconfig-stage.oeclass
@@ -82,6 +82,9 @@ def binconfig_stage_fixup(d):
             dstlink = os.path.join(
                 stage_dir, cross_type, d.get("stage_bindir").lstrip("/"),
                 os.path.basename(filename))
+            dstdir = os.path.dirname(dstlink)
+            print("creating destination dir: {}".format(dstdir))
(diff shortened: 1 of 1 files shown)


hundeboll:
I created to recipes to test this:
https://gist.github.com/01f7bce90cb46923f6ba868602925494
https://gist.github.com/555241212581edb7baa29b389eb3b0cb

To avoid building the toolchain, package-qa should be removed from package.oeclass:
https://gist.github.com/61f633e004c5fbc3961c54d5cb828ac0

esben:
As discussed on gitter, messing with stage dir directly was never supposed to be done in stage fixup functions.  But I don't see any other easy fix here and now, so I guess we will have to live with it for now.

PR #141, opened 2016-06-07
classes/binconfig: create binconfig package instead of doing binconfig-fixup

hundeboll:
Packages that create a ${PN}-config script used by dependent packages
during their configure step, inherits the binconfig class to both fixup
paths in the script and to make the script available to dependent
packages during their stage step.

Until now, this was done by the binconfig class, which adds a call to
binconfig_stage_fixup() during the dependent stage step. However,
binconfig_stage_fixup() symlinks into ./stage/cross/bin/, which is
not "allowed" in the stage fixup functions. (They should only touch
./stage.unpack/ and not ./stage ).

Also, the symlink-during-stage fixup approach might fail as
./stage/cross/bin is not guarenteed to be created when
binconfig_stage_fixup() is called.

Fix this mess by changing the class to create the symlink during
do_install(), and adding the symlink to cross-package created in the
class (${PN}-binconfig). Provide this package to dependent packages
by adding cross:${PN}-binconfig as a dependency to the primary package.


esben:
You are also dropping all the path rewrite hackery.  Could you please add some comments on why this is a good thing?

hundeboll:
My mistanke; i'll resubmit tomorrow...

classes/binconfig.oeclass
```diff
@@ -1,13 +1,34 @@
...
+##                        files (default to "${HOST_SYSROOT} ${TARGET_SYSROOT} ${D}
+##                        ${B} ${S}")
 
 require conf/meta.conf
 
 BINCONFIG_FILES ?= "${bindir}/*-config"
 BINCONFIG_GLOB ?= "${BINCONFIG_FILES}"
 FILES_${PN}-dev += "${BINCONFIG_FILES}"
 BINCONFIG_FIXUP_STRIP_DIRS ?= "${HOST_SYSROOT} ${TARGET_SYSROOT} ${D} ${B} ${S}"
+BINCONFIG_PACKAGE ?= "${PN}-binconfig"
```
esben: Maybe call it ${PN}-binconfig-cross to make it more clear that it is not actually the binconfig files, but just a cross thingy  (ie. symlink) of it.

classes/binconfig.oeclass
```diff
@@ -1,13 +1,34 @@
...
+BINCONFIG_LIB ?= "${PN}"
```
esben: More naming.  It seems weird to call it BINCONFIG_LIB when it is just the name of a package containing one (or more) binconfig scripts.

Maybe BINCONFIG_CROSS_PACKAGE and BINCONFIG_PACKAGE instead of BINCONFIG_PACKAGE and BINCONFIG_LIB.

classes/binconfig.oeclass
```diff
@@ -45,6 +66,15 @@ def do_install_binconfig_fixup(d):
...
             output_file.write(binconfig_file)
 
+        # create symlink
+        srcdir = os.path.join("..", "..", "machine", d.get("machine_bindir").lstrip("/"))
+        srcfile = os.path.join(srcdir, os.path.basename(filename))
+        dstdir = os.path.join(d.get("D"), d.get("stage_bindir").lstrip("/"))
+        dstlink = os.path.join(dstdir, os.path.basename(filename))
+        oelite.util.makedirs(dstdir)
+        print "symlinking %s to %s"%(dstlink, srcfile)
+        os.symlink(srcfile, dstlink)
```
esben: What happens if ${machine_bindir} == ${stage_bindir} ?

mnhu: Nothing good will come from that.. We might need a better way to separate the symlink from the real script. Does it make sense to rename the real script prior to creating the symlink?

mnhu: The same issue arises in line 30:
`FILES_${BINCONFIG_CROSS_PACKAGE} = "${stage_bindir}/*-config"`

PR #144, opened 2016-06-14
classes/manifest-version: include unannotated tags and abbrev commit id

hundeboll:
Using unannotated tags leads to "wrong" manifest versions (i.e. the last
annotated tags instead of the latest unannotated tag).

Including the abbrev commit id makes it possible to backtrack a commit if the
tag is overwritten.

### DIFF
--- a/classes/manifest-version.oeclass
+++ b/classes/manifest-version.oeclass
@@ -13,7 +13,7 @@ def manifest_version(d):
     os.chdir(d.get("TOPDIR"))
     version = None
     if os.path.exists(".git"):
-        cmd = subprocess.Popen(["git", "describe", "--dirty", "--always"],
+        cmd = subprocess.Popen(["git", "describe", "--dirty", "--always", "--tags"],
                                stdout=subprocess.PIPE, stderr=subprocess.PIPE)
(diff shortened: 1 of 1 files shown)


esben:
I would prefer to keep manifest-version only supporting annotated tags.  I don't really see the point in using unannotated tags for anything but private/personal use, ie. remembering a specific commit.

Anything worthy of being considered part of the manifest version should be annotated.
See fx. http://stackoverflow.com/questions/4971746/why-should-i-care-about-lightweight-vs-annotated-tags for a discussion of annotated vs unannotated tags.

As for overwriting tags.  Don't do that!  You will suffer permanent pain, as you have no way of ensuring that everybody uses the same version of the tag.  Adding support for this in OE-lite gains nothing but a false assumption that you will be able to handle it.

That said, it might be a matter of personal taste to use `--long`, so I won't mind adding support for that in manifest-version.  But it should not be the default, and it will not be in there to support handling of overwritten tags (as far as I am concerned).

hundeboll:
This is no longer needed due to #218

esben:
Thanks :)

PR #146, opened 2016-06-16
auto-package-libs.oeclass: allow configuring libprefix

Villemoes:
Currently, one could achieve rougly the same thing by overriding
AUTO_PACKAGE_LIBS_LIBDIR, setting it to e.g. "${libdir}:myprefix
${sharedlibdir}:myprefix", but it's a lot more convenient to be able
to set the default prefix independently from the libdirs.

The use case is e.g. libunwind, which generates libraries
libunwind.so, libunwind-setjmp.so, libunwind-ptrace.so,
libunwind-coredump.so etc. Rather than making AUTO_PACKAGE_LIBS
contain "unwind-setjmp unwind-ptrace" with corresponding packages
"libunwind-libunwind-setjmp" etc, it's more convenient to set both
PROVIDEPREFIX and LIBPREFIX to "${PN}-" and PKG_PREFIX to "", and then
just list the shorter names "setjmp ptrace" etc. in
AUTO_PACKAGE_LIBS. That way, we get a package "libunwind-ptrace" which
provides "libunwind-ptrace.so".

Signed-off-by: Rasmus Villemoes <rasmus.villemoes@prevas.dk>

### DIFF
--- a/classes/auto-package-libs.oeclass
+++ b/classes/auto-package-libs.oeclass
@@ -33,6 +33,7 @@ AUTO_PACKAGE_LIBS ?= ""
 AUTO_PACKAGE_LIBS_LIBDIR ?= "${libdir} ${sharedlibdir}"
 AUTO_PACKAGE_LIBS_PKGPREFIX ?= "lib"
 AUTO_PACKAGE_LIBS_PROVIDEPREFIX ?= "lib"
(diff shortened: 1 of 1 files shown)


esben:
Good idea :+1:

PR #147, opened 2016-07-11
Ravi/dynvarstuff

Villemoes:
Following discussion on gitter, here's a few patches that affect the handling of "dynvars" and, importantly, change the hash value of every task. I suggest merging it as one of the first towards 7.0.0.

To check that I've caught all the variables affected by the changed dynvar logic, here's what I did:

```
cd meta/core && git checkout master && cd ../..
oe bake universe --dryrun --dump-signature-metadata=tmp/meta_before
cd meta/core && git checkout ravi/dynvarstuff && cd ../..
oe bake universe --dryrun --dump-signature-metadata=tmp/meta_after
# remove variables we know are gone from the hash computation
find tmp/meta_before/ -type f | xargs sed -i '/^FILE_DIRNAME=/d'
find tmp/meta_before/ -type f | xargs sed -i '/^SIGNATURE_FILE=/d'
find tmp/meta_before/ -type f | xargs sed -i '/^FILESPATH_EXISTS=/d'
# check that the only difference is now the extra newlines from the last patch
diff -B -u -r tmp/meta_before/ tmp/meta_after/
# (no output)
```


lib/oelite/baker.py
```diff
@@ -250,7 +250,7 @@ def show(self):
 
         #meta.dump(pretty=False, nohash=False, flags=True,
         #          ignore_flags=re.compile("filename|lineno"),
-        meta.dump(pretty=True, nohash=(not self.options.nohash),
+        meta.dump(pretty=True, show_nohash=(not self.options.nohash),
```
kimrhh: i don't really want to diverge your work, so maybe this should just be ignored since the change is ok as long as you accept the general quality of the current code, but

Methods with boolean args is a CodeSmell, and it is a strong indication that this should be refactored into something else interely. Why not a seperate method dumpVerbose() perhaps?

lib/oelite/baker.py
```diff
@@ -76,9 +76,12 @@ def add_bake_parser_options(parser):
 
 
 def add_show_parser_options(parser):
+    import optparse
     parser.add_option("--nohash",
                       action="store_true",
                       help="don't show variables that will be ignored when computing data hash")
+    parser.add_option("--nodynvar",
+                      action="store_true", help=optparse.SUPPRESS_HELP)
```
kimrhh: I don't think it should be hidden, but I also fail to understand how this is useful. It seems to me that you are being confused about the output showing "not fully expanded" stuff, but are implementing a handle to expand even less?

suggestion for help text:
help="Do not show the expanded value of the special 'dynamic variables' which are values the oe-lite generates based on e.g. git checkout information"

lib/oelite/baker.py
```diff
@@ -251,6 +254,7 @@ def show(self):
         #meta.dump(pretty=False, nohash=False, flags=True,
         #          ignore_flags=re.compile("filename|lineno"),
         meta.dump(pretty=True, show_nohash=(not self.options.nohash),
+                  dynvar_replacement=(not self.options.nodynvar),
```
kimrhh: ugh more of this boolean arg stuff, it is really not making the code nicer to work with :)

lib/oelite/meta/meta.py
```diff
@@ -436,15 +436,18 @@ def dump_var(self, key, o=sys.__stdout__, pretty=True, dynvars=[],
...
 
         val = str(val)
 
-        for dynvar_name, dynvar_val in dynvars:
-            val = string.replace(val, dynvar_val, dynvar_name)
+        if deps:
+            for dynvar_name, dynvar_token, dynvar_val in dynvars:
+                if dynvar_name not in deps:
+                    continue
+                val = string.replace(val, dynvar_val, dynvar_token)
```
kimrhh: am I reading this right, that with this change e.g. SIGNATURE_FILE will no longer become a part of the hash? if so, that does not seam like a good idea :)

kimrhh: ok, thinking some more about it, it would be a pretty weird case if you had to change the the path to the .sig file. Grepping for use of the other PATH variables mentioned above, it seems like the only normal manipulation FILEPATH variables is the FILEPATHPKG. so I guess those are also safe to nohash

lib/oelite/meta/meta.py
```diff
@@ -462,13 +457,13 @@ def dump_var(self, key, o=sys.__stdout__, pretty=True, dynvars=[],
             return
 
         if func == "bash":
-            o.write("%s() {\n%s}%s"%(key, val, eol))
+            o.write("%s() {\n%s}\n\n"%(key, val))
```
kimrhh: speeeeling error in commit headline

kimrhh:
Nice work! overall this seems mostly ready IMO.

lib/oelite/baker.py
Villemoes: No, this doesn't expand less. It prevents "unexpanding", so in that sense it expands more (e.g., showing DATETIME="20160102030405" instead of DATETIME="${DATETIME}"). And yes, I've occasionally been confused by the not fully expanded stuff, but now I understand why it happens, so I'm not going to push very hard for this option to get in if others find it entirely useless.

lib/oelite/meta/meta.py
Villemoes: As I write in the commit message, having the string '${TOPDIR}/relative/path/recipe.sig' mixed into the hash doesn't really provide anything which isn't already accounted for by other variables which do get hashed. Moreover, even if somebody decides to do 'SIGNATURE_FILE = "${PN}.something_else_just_because_im_special", what is really important for hashing is the signatures contained in that file.

The patch adding these variables to builtin_nohash is a necessary followup to the "safer/saner" dynvar handling.

lib/oelite/baker.py
esben: I think it is useful.

esben: Please add a help text.  I don't want this option hidden.

esben:
Other than that: LGTM

Villemoes:
Added help text and rebased to current master (i.e., post 6.3.0). Please wait a little with merging, however - I'd like to do a little more investigation of the effect on metadata hashes - in particular, to see if there are more variables that needs to be handled differently now.

PR #148, opened 2016-07-11
remove sqlite use, baby steps

Villemoes:
The most significant part of these patches is that they cut "building dependency tree" for oe bake world from 1m39s to about 18s. Aside from that, the diffstat is also quite nice; about 100 lines removed from files that are still used.

One could certainly argue that abe0274 and d59ab1a should be squashed, but I'd like to keep a record of how I got there.


lib/oelite/runq.py
```diff
@@ -352,7 +352,7 @@ def resolve_dependency(item, recursion_path, deptype):
             recursion_path[1].append(str(item))
 
             # try to get cached recdepends list of packages
-            packages = self.get_recdepends(package, deptype)
```
kimrhh: shouldn't the get_recdepends method in runq die then?

kimrhh:
looks good to me, if we can actually demonstrate it does not make the builds splat :)

lib/oelite/runq.py
Villemoes: Isn't that what e173380 does?

kimrhh: oh, indeed. :+1:

PR #149, opened 2016-07-13
oelite: add check-provided subcommand

Villemoes:
This is a quick'n'dirty proof-of-concept, but I think it might already be useful in its current form, and it shouldn't really affect normal operation of the oe command, hence the PR.


lib/oelite/cmd/check_provided.py
```diff
@@ -0,0 +1,182 @@
...
+data["mercurial"] = dict(cmd = "hg")
+data["mtd-utils-mkfs-jffs2"] = dict(cmd = "mkfs.jffs2", ret = 255)
+data["perl"] = dict(extract = r"\(v([0-9][0-9.a-zA-Z_-]+)\)")
+data["pkgconfig"] = dict(cmd = "pkg-config")
+data["python-runtime"] = dict(cmd = "python")
+data["texinfo"] = dict(cmd = "texindex")
+data["unzip"] = dict(args = "-v")
+data["util-linux"] = dict(cmd = "getopt")
+
+data["git"] = dict(min = "1.8.4")
```
kimrhh: it doesn't seem like a good idea to add version information here, since I don't think everybody would know they need to update a sub-command if the touch provided.conf :)

Would it be a problem to add this information to the .conf file instead?

kimrhh:
nice addition, but think a command should not contain any data, since this is confusing from a user perspective

lib/oelite/cmd/show_provided.py
```diff
@@ -0,0 +1,185 @@
...
+        if u in data:
+            data[u][key] = v
+
+def get_provided_version(p):
+    param = data.get(p)
+    if not param:
+        return None
+
+    args = [param["cmd"]] + param["args"]
+    cmdstring = " ".join(args)
```
esben: Why are you flattening args?  subprocess.Popen() is capable of taken args as a sequence, and in case you have args with whitespace, you effectively split the arguments this way.

Villemoes: Please look at the Popen call - I do pass args as a list. cmdstring is only used in warning messages. Of course that could end up being slightly confusing if there are arguments with whitespace, but I really don't want to complicate this with trying to get shell quoting right.

PR #150, opened 2016-08-02
RFC: run tasks in parallel

Villemoes:
This is not ready for merging, but I want to get some comments before spending time on the necessary wider testing, cleanup and "hardening" (testing with various errors injected deliberately).

The last commit shows the potential: when building multiple things, we can get stuff done about 40% faster (wallclock time), and that's before trying to be smart. One of the side effects of this series is what I've wanted for a long time: a breakdown of the time spent on each individual task, and that shows what I've long suspected: do_configure is rather time-consuming, since it usually runs a gazillion compile processes completely serialized.

The overall strategy is quite simple: Instead of having a single task.run() method, split that into a method that starts the task and a method for waiting (or just polling) for its completion, keeping the list of currently running tasks in a suitable data structure. However, the concept of pre- and postfuncs complicates that strategy somewhat, as does the fact that the actual shell script gets called four layers down (baker.py -> task.py -> function.py -> util.shcmd). I took a bottom-up approach, first getting rid of the last layer, since I need to have control over the spawning and waiting for of the subprocess. That also allows me to keep things working as usual by implementing run() as start()+wait() in each of the layers.

### DIFF
--- a/lib/oelite/baker.py
+++ b/lib/oelite/baker.py
@@ -577,7 +577,7 @@ class OEliteBaker:
             count += 1
             debug("")
             debug("Preparing %s"%(task))
-            task.prepare(self.runq)
+            task.prepare()
             meta = task.meta()
             info("Running %d / %d %s"%(count, total, task))
             task.build_started()
--- a/lib/oelite/baker.py
+++ b/lib/oelite/baker.py
@@ -576,10 +576,17 @@ class OEliteBaker:
 
         oven = OEliteOven(self, 8)
         try:
-            while True:
+            while oven.count < oven.total:
                 pending += self.runq.get_runabletasks()
-                if not pending:
-                    break
+                if not pending or oven.capacity <= 0:
+                    # If we have no runable tasks and nothing in the
+                    # oven, some tasks must have failed.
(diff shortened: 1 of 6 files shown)


Villemoes:
I forgot to mention: I'm aware that this more or less implements Yet Another Asynchronous Framework of which I'm sure there are dozens, but that's actually a lot easier than trying to graft something existing onto oe-lite, with our "weird" mix of python and shell functions. So yes, it's probably not very elegant, and yes, this is not how it would have looked if we were designing oe-lite from scratch with parallel tasks in mind.

Villemoes:
After fixing two recipes in current base I got "oe bake world" to succeed, and thankfully that's also the case with this series. The ~40% is still true:

Current master:

rm -rf tmp/{cache,stamp,work}
time oe bake world -y

Build time 1h19m32s

real    81m52.882s
user    166m47.584s
sys     19m47.348s

With parallel tasks enabled:

rm -rf tmp/{cache,stamp,work}
time oe bake world -y

Build time 49m04s

real    51m25.271s
user    208m51.440s
sys     26m30.992s

PR #154, opened 2016-08-24
passwd classes: make them more robust

Villemoes:
There's a bug in the current avahi recipe: The autoipd-passwd and
daemon-passwd files end up being packaged under the names avahi.1 and
avahi.2, but in the packages' respective FILES_* list they list each
others' files. This means that when one includes one but not the other
package, the expected user account is missing from the system.

Having to manually enumerate the PASSWD_FILES the same way the passwd
class does it is rather fragile. So instead of renaming the files to
${PN}.${i}, put each recipe's passwd files in the subdirectory ${PN} and
keep the file's basename. That makes the relevant FILES entries much
more predictable.

While at it, replace $cwd with . in image_passed.oeclass - avoiding
long absolute paths makes it easier to read the log output.

### DIFF
--- a/classes/passwd.oeclass
+++ b/classes/passwd.oeclass
@@ -1,11 +1,20 @@
 ## OE-lite class for recipes providing content for /etc/passwd and /etc/group
 ##
-## If a recipe has contents for /etc/passwd or /etc/group, just ensure that the
-## content for /etc/passwd is placed in PASSWD_FILES, and contents for
-## /etc/group in GROUP_FILES. The contents from a recipe is placed in
-## a recipe specific dir, so that multiple recipes may provides contents for
-## /etc/passwd and /etc/group. The content is later combined in the
-## image_passwd class.
+## If a recipe has contents for /etc/passwd or /etc/group, just ensure
+## that the content for /etc/passwd is placed in PASSWD_FILES, and
+## contents for /etc/group in GROUP_FILES. The contents from a recipe
+## is placed in a recipe specific dir, so that multiple recipes may
+## provides contents for /etc/passwd and /etc/group. The content is
+## later combined in the image_passwd class.
+##
+## Currently, all files from a given recipe must have different
+## basenames; the file $f gets copied to
+##
+##   ${passwddir}/${PN}/$(basename $f)
+##
+## Should this ever become a problem that cannot be solved by
+## appropriate renaming, see https://github.com/oe-lite/core/pull/154
+## for a possible extension.
 ##
 ## @var PASSWD_FILES Set the ${SRCDIR}/passwd by default.
 ## @var GROUP_FILES Set to ${SRCDIR}/group by default.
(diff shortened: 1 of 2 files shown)


classes/passwd.oeclass
```diff
@@ -19,18 +19,14 @@ GROUP_FILES ?= "${SRCDIR}/group"
 PASSWD_POSTFUNC:machine = " do_install_passwd"
 do_install[postfuncs] += "${PASSWD_POSTFUNC}"
 do_install_passwd () {
-	i=0
 	for f in ${PASSWD_FILES} ; do
-		mkdir -p ${D}/${passwddir}
-		let i=$i+1
-		cp $f ${D}/${passwddir}/${PN}.$i
+		mkdir -p ${D}/${passwddir}/${PN}
+		cp $f ${D}/${passwddir}/${PN}/
```
esben: This would not end very well with something like

`PASSWD_FILES="${SRCDIR}/foo/my-file ${SRCDIR}/bar/my-file`

esben:
A side effect of this change is that ordering of the lines provided by multiple files in `${PASSWD_FILES}` is not preserved.  It is probably only of cosmetic interest though.

Villemoes:
Huh? The final /etc/passwd file is explicitly sorted by uid, so there's no implicit guarantee of ordering.

classes/passwd.oeclass
Villemoes: That's true. Could we make ${SRCDIR} implicit, so one should say PASSWD_FILES="foo/my-file bar/my-file", and then copy to the corresponding path under ${PN}? And instead of globbing we'd just do 'find ${passwddir} -type f | xargs cat >> ${passwdfile}'. Adding -n to cp is probably a good idea in any case.

esben:
Ah, sorry.  Forgot about that :)

classes/passwd.oeclass
esben: 2 counter arguments to that:

1. That would be different from most other similar interfaces, thus violating principle of least surprise.
2. That would prevent the use of files from other locations, such as ${B} and ${STAGE_DIR}.

Villemoes: Agreed; (2) also occured to me after I posted this. But I'd still like to find a way to name and place the passwd files under ${passwddir} in a completely predictable manner so it's easier to write correct FILES variables. In particular, it should be possible to do PASSWD_FILES += "foo-passwd", FILES_${PN}-foo += "some_function_of(foo-passwd)" anywhere in a recipe without having to worry about what other entities PASSWD_FILES contained at that point.

Villemoes: How about this: We place the passwd files from a given recipe under ${passwd}/${PN}/ as above. We use cp -n to cause an error if two files would collide. If for some reason one has two files with the same basename and that can't easily be changed, a PASSWD_FILES entry can have an optional ":where/to/put/it" appended. So your example would be `PASSWD_FILES = "${SRCDIR}/foo/my-file:foo/my-file ${SRCDIR}/bar/my-file:bar/my-file"`. We already have a few places where we pass multiple pieces of information as a colon seperated list (e.g. some AUTO_PACKAGE_LIBS vars), and rely on nobody sane putting a : in a path. No current recipes need this AFAICT, and when the passwd files come from a file:// it's probably always easier to just follow the avahi example and use ${packagename}-passwd,

esben: I completely agree with the intent of this PR.

And your latest suggestion sounds fine.  I don't even think we have to do anything special to handle conflicting files (other than detect it and bail out).  If/when it happens, it should be pretty trivial to rename the file by normal means.  Adding extra complexity for that is probably overkill.

Villemoes: Well, if one is taking files from $B one might not have that much control over their basenames, so it would require somewhat more than a rename. But no-one does that right now, so I'll drop the complexity and just add -n, and maybe a link to this very PR so we can revisit should the need arise.

esben: I were thinking it would be easy to add a postfunc which renamed the files.

PR #157, opened 2016-08-31
auto_package_utils.oeclass: add AUTO_PACKAGE_UTILS_SYMLINKS_* feature

Villemoes:
Some binaries change behaviour based on argv[0], and those utilities
often come with a bunch of symlinks pointing to the same binary. With
the current auto_package_utils, these symlinks usually end up as
dangling links in the main package with no tie to the actual package
providing the binary.

Adding the symlink names to the AUTO_PACKAGE_UTILS var is not really a
solution, as that would just create new packages containing dangling
links.

One could do something like

FILES_xz-xz += "/usr/bin/lzcat"
FILES_xz-xz += "/usr/bin/lzma"
FILES_xz-xz += "/usr/bin/unlzma"
FILES_xz-xz += "/usr/bin/unxz"
FILES_xz-xz += "/usr/bin/xzcat"

but that doesn't tell the world that the xz-xz package actually
provides util/lzma.

So this implements support for spelling the above

AUTO_PACKAGE_UTILS_SYMLINKS_xz = "lzcat lzma unlzma unxz xzcat"

which will, in addition to putting the symlinks in the xz-xz package,
also make sure that xz-xz provides util/lzma etc.

Tested with the (not-yet-merged) xz recipe by looking at what ends up
in the packages and the value of the PROVIDES_ variables.

### DIFF
--- a/classes/auto-package-utils.oeclass
+++ b/classes/auto-package-utils.oeclass
@@ -18,6 +18,13 @@
 ##      Defaults to PN.
 ## @var AUTO_PACKAGE_UTILS_PREFIX Prefix used when looking up files to
 ##      package.  Defaults is empty string.
+## @var AUTO_PACKAGE_UTILS_SYMLINKS_foo List of alternate names for
+##      the "foo" utility, usually implemented via symlinks. This will
+##      put these symbolic links into the ${PN}-foo package, and that
+##      package will automatically provide "util/bar", "util/baz",
+##      ... in addition to "util/foo". For example,
+##
+##      AUTO_PACKAGE_UTILS_SYMLINKS_bzip2 = "bunzip2 bzcat"
 
 PACKAGES =+ "${AUTO_PACKAGE_UTILS_DOC_PACKAGES}"
 PACKAGES =+ "${AUTO_PACKAGE_UTILS_PACKAGES}"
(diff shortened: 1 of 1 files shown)


esben:
Please add description of how to use AUTO_PACKAGE_UTILS_SYMLINKS to top of auto-package-utils.oeclass file.

classes/auto-package-utils.oeclass
```diff
@@ -63,24 +63,36 @@ def auto_package_utils (d):
...
+                      "${base_bindir}/%s "%(b) +
+                      "${sbindir}/%s "%(b) +
+                      "${bindir}/%s "%(b) +
+                      "${libexecdir}/%s "%(b))
+            if exeext:
+                files += (" ${base_sbindir}/%s%s "%(b, exeext) +
+                          "${base_bindir}/%s%s "%(b, exeext) +
+                          "${sbindir}/%s%s "%(b, exeext) +
+                          "${bindir}/%s%s "%(b, exeext) +
+                          "${libexecdir}/%s%s "%(b, exeext))
```
esben: So, on Windows you will include both with and without .exe, why not just the .exe files?  The other will not be seen as executable anyway, and will therefore were likely never exist.

Villemoes: no idea, just copied what was already there for the util itself. Does windows even have the symlink concept?

esben: Sorry, didn't see that you were already done that way.  In that case, a fix for that does not really belong here.  Carry on :)

As for Windows and symlinks.  For all practical purposes, no, Windows does not have symlinks.

esben:
Please fix the build, so we can get this merged for 6.1.0.

ravi-prevas:
Who is that directed at? I don't think this is responsible for "FATAL: No provider for native:python3-freeze-importlib (needed by recipe machine:python_3.5.1)".

esben:
I don't know who it is directed at.  I guess whoever knows why python keeps breaking builds for non-python related changes...

esben:
The build might be fixed with oe-lite/python#4

esben:
Uhu, green lights :)

PR #161, opened 2016-09-16
bake.py: add warning if building on unsupported os

vmbracvlvm:
Print a warning if building on unsupported os. The warning triggers a
prompt asking to continue or abort build. Use '-s' (--skip-os-check) option to
automatically continue build.

To add a supported OS see core/setup directory. Files can also be added in
a setup folder in local meta layer. To unsupport an OS add a
setup file with only the distro-release line (keep only the first line).

It would be handy to have a feature to disable all other setup folders
but the one in local layer. But so far this has not been implemented.


lib/oelite/cmd/bake.py
```diff
@@ -23,8 +28,34 @@ def parse_args(options, args):
...
+                        print
+                        response = 'n'
+                    response = response.lower()
+                    if response == '':
+                        response = 'y'
+                    if response in ('y', 'n'):
+                        print
+                        break
+
+            if response == 'n':
```
esben: Won't this break when stdin is not tty?  Maybe just initialize response to 'n' in an else clause to the tty if.

esben:
Any progress on this?

vmbracvlvm:
Setting initial value of response to 'n' thus failing per default. Use -s to force continue.

lib/oelite/cmd/bake.py
esben: Better :)

esben:
Before merging this, we really need to have a setup file for Ubuntu 16.04 (and probably 15.10 as well).

PrevasJsch:
Im on it

vmbracvlvm:
setup files has been vaguely tested in official ubuntu containers

esben:
Oh, if you add a debian_8 setup file, build should hopefully go green.

vmbracvlvm:
Still need to detect debian distro in setup.get_host_distro()

lib/oelite/cmd/bake.py
```diff
@@ -23,8 +28,46 @@ def parse_args(options, args):
...
+    distro, release = setup.get_host_distro()
+    if distro and release:
+        setup_file, supported = setup.get_setup_file(config, distro, release)
+        if supported:
+            return True
+
+    if distro and release:
+        print('WARNING: Host distribution (%s %s) has not been validated'%(distro, release))
+    else:
+        print("WARNING: Cannot determine host distribution")
```
kimrhh: this stuff does not really belong in bake. Let us make good use of OO and keep the responsibility for the logic about support host OS in the setup code.

What is needed is a refactor of the setup code, so you can simply ask it from bake "are we running on a supported host OS", probably something like a "is_supported()" method or the like. The rest of the setup code should of cause use that as well

lib/oelite/cmd/bake.py
```diff
@@ -23,8 +28,46 @@ def parse_args(options, args):
...
+            if response == '':
+                response = 'y'
+            if response in ('y', 'n'):
+                print
+                break
+
+    if response == 'n':
+        print "Maybe another time"
+        return False
+    return True
```
kimrhh: isn't this a duplication of the other place where we interactively ask if we should continue?

Let us not do that, we already have a mess of duplication and responsibility in the code. Please refactor this out to a place where it can be used by commands that need it.

lib/oelite/cmd/bake.py
```diff
@@ -23,8 +28,46 @@ def parse_args(options, args):
...
 
 def run(options, args, config):
+
+    if not options.skip_os_check:
+        if not os_check(config):
```
kimrhh: this line is basically what I would imagine should be needed in bake, but by calling into the setup code.

esben:
@vmbracvlvm @kimrhh Who is going to follow up on this?

kimrhh:
I will look at it some time, but don't know when

esben:
Ok, it will have to be delayed to 7.0 then.

PR #166, opened 2016-09-16
fetch/git.py: write to log when branch, rev or tag not found

vmbracvlvm:
When a recipe try to use git fetch, with a revision or tag that does not exist in that checkout, the fetch will fail without any information about why it failed.


esben:
LGTM.

PR #168, opened 2016-09-25
Parallel task preparation

Villemoes:
These patches should be completely innocent. Some remove dead code (which is nice to get rid of, because that makes it easier to reason about the patches that implement the parallel task feature). Some just do some trivial refactoring. Some prepare for performance/correctness improvements which are not directly parallel task-relevant. And some add profiling infrastructure, since I was getting tired of having to bolt that on and off when trying to measure things. The profiling which is done by default obviously has to be extremely low overhead, but having the rest of the infrastructure available means I only have to add a single line or two instead of the entire implementation.

From here, it's only about 15 more patches to parallel tasks :)

### DIFF
--- a/lib/oelite/compat.py
+++ b/lib/oelite/compat.py
@@ -1,0 +1,82 @@
+#!/usr/bin/python
+import os
+import fcntl
+import ctypes
+import datetime
+import sys
+
+__all__ = ["dup_cloexec", "open_cloexec"]
+
+uname = os.uname()
+sysname, release, machine = (uname[0], uname[2], uname[4])
+
+values = dict()
+if (sysname, machine) == ("Linux", "x86_64"):
+    # these are supported since at least 2.6.24, so just set them unconditionally
+    values['O_CLOEXEC'] = 02000000
+    values['F_DUPFD_CLOEXEC'] = 1024+6
+
+# import bb.utils doesn't work when we try to run this by itself (to
+# test it), so this just serves to show what one could do, provided
+# someone figures out how to make the vercmp_string available.
+#
+# elif sysname == "FreeBSD":
+#     if bb.utils.vercmp_string(release, "8.3") >= 0:
+#         values['O_CLOEXEC'] = 0x00100000
+#     if bb.utils.vercmp_string(release, "9.2") >= 0:
+#         values['F_DUPFD_CLOEXEC'] = 17
+
+def dup_cloexec(fd):
+    return fcntl.fcntl(fd, F_DUPFD_CLOEXEC, 0)
+
+def open_cloexec(filename, flag, mode=0777):
+    return os.open(filename, flag | O_CLOEXEC, mode)
+
+def set_cloexec(fd):
+    fcntl.fcntl(fd, fcntl.F_SETFD, fcntl.FD_CLOEXEC)
+    return fd
+
+def dup_cloexec_fallback(fd):
+    return set_cloexec(os.dup(fd))
+
+def open_cloexec_fallback(filename, flag, mode=0777):
+    return set_cloexec(os.open(filename, flag, mode))
+
+if hasattr(os, "O_CLOEXEC"):
+    O_CLOEXEC = os.O_CLOEXEC
+else:
+    try:
+        O_CLOEXEC = values["O_CLOEXEC"]
+    except KeyError:
+        open_cloexec = open_cloexec_fallback
+
+if hasattr(fcntl, "F_DUPFD_CLOEXEC"):
+    F_DUPFD_CLOEXEC = fcntl.F_DUPFD_CLOEXEC
+else:
+    try:
+        F_DUPFD_CLOEXEC = values["F_DUPFD_CLOEXEC"]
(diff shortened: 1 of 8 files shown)


lib/oelite/profiling.py
```diff
@@ -98,9 +97,10 @@ def print_deferred(self):
         for p in self.deferred:
             p.print_delta()
 
-    def __init__(self, name):
+    def __init__(self, name, print_walltime=True):
```
kimrhh: you are changing profiling.py in a commit that only mentions baker.py. Shouldn't it be split up?

Villemoes: I suppose I could do the teaching in a separate commit, but while the commit message doesn't mention profiling.py by that name, it does refer to oelite.profiling.

Villemoes:
Superseded by #171 .

lib/oelite/util.py
```diff
@@ -57,6 +57,11 @@ def write(self, text):
             file.write(text)
         return
 
+def stracehack(msg):
```
mnhu: shouldn't this be enabled by some global flag? It might confuse unaware users should they happen to debug oe-lite with strace too...

lib/oelite/compat.py
```diff
@@ -0,0 +1,82 @@
...
+import os
+import fcntl
+import ctypes
+import datetime
+import sys
+
+__all__ = ["dup_cloexec", "open_cloexec"]
+
+uname = os.uname()
+sysname, release, machine = (uname[0], uname[2], uname[4])
```
mnhu: why not one line with `sysname, release, machine = os.uname()` ?

lib/oelite/profiling.py
```diff
@@ -0,0 +1,188 @@
...
+            return somefunc(*args, **kwargs)
+        finally:
+            delta = (now()-start).total_seconds()
+            if not somefunc in profiled_functions:
+                profiled_functions[somefunc] = SimpleStats()
+            profiled_functions[somefunc].append(delta)
+    return recordtime
+
+def write_call_stats():
+    outfn = os.path.join(profiledir, "call_stats.txt")
```
mnhu: profiledir can be `None` if the user hasn't called init() in advance...

lib/oelite/profiling.py
```diff
@@ -0,0 +1,188 @@
...
+                srcfile = os.path.basename(inspect.getsourcefile(f))
+            except TypeError:
+                srcfile = "<unknown>"
+
+            stats = profiled_functions[f]
+            stats.compute()
+            out.write("%-12s\t%-24s\t%9.3fs / %5d = %7.3f " %
+                    (srcfile, f.__name__,
+                     stats.sum, stats.count, stats.mean))
+            out.write("[%s]\n" % ", ".join(["%7.3f" % x for x in stats.quartiles]))
```
mnhu: jeeez, please use placeholders instead of doing list comprehension as a function arguement...

esben:
Oh no, was it really necessary to close this PR in favour for the much bigger oe-lite/core#171?

Villemoes:
I wanted to give people a chance to  comment on the much more important patches that actually lead to parallel tasks - for example, as kimrhh has noted, I'm slaying some existing functionality to get where I want. In hindsight, the "prepare PR" model really sucks, partly because I have to maintain two branches in sync, partly because reviewers don't have easy access to "where are we going with this", and partly exactly because the review comments for the actual series gets split across at least two PRs. Splitting the head of one branch off and designating that the prep branch also makes it much more cumbersome to move patches from or to the prep status.

I will make sure to take all comments from here into account.

lib/oelite/util.py
Villemoes: Believe me, when you strace oe-lite there are many many more and much larger WTF moments than meeting an occasional odd write (for example one might wonder where stat("") comes from and what good that could ever do). And having a way to isolate where the odd (or perhaps just excessively many) syscalls come from is exactly what this is for. Also, "git grep -E '[-]42'" and other semi-obvious grep-patterns immediately lead to util.py, and then an explanation is just a git blame away.

esben:
Ok.

lib/oelite/compat.py
Villemoes: Well, at least it would have to be `sysname, _, release, _, machine = os.uname() ` to throw away `uname[1]` and `uname[3]`. That's a standard python idiom, but I don't write enough snake to think of writing stuff that way, and for the same reason I also have to think twice when I read code like that.

lib/oelite/profiling.py
Villemoes: True, but (a) write_call_stats is only called as an atexit handler, which is only installed in init(), and (b) commit b4d235b (oelite/profiling.py: implement profile_output function) removes this problem altogether. I probably should squash that into the original commit creating profiling.py.

lib/oelite/compat.py
Villemoes: OK, apparently I'm lying - I'm using that very idiom in profiling.py...

PR #169, opened 2016-09-30
linux-headers: version bump to 4.4.23

sknsean:
Bump linux-headers to newest LTS


sknsean:
Depends on https://github.com/oe-lite/base/pull/394

kimrhh:
are we just going to let anyone downstream with kernels <=4.4 "keep the pieces" or could we add a big fat warning if they are ending up with a combination that is not good?

any real reason to delete the old headers?

esben:
How does this depend on oe-lite/base#394 (socat upgrade)?

sknsean:
Can't remeber :-)

PR #170, opened 2016-10-04
setup/ubuntu*: generate en_US.UTF-8 locale

ravi-prevas:
The en_US.UTF-8 locale is not generated/installed by default. But since
we have

LANG ?= "en_US.UTF-8"

in conf/shell.conf, any recipe using perl (e.g. linux-headers) ends up
spewing lots of stuff like below to the log.

perl: warning: Setting locale failed.
perl: warning: Please check that your locale settings:
        LANGUAGE = (unset),
        LC_ALL = (unset),
        LC_NUMERIC = "C",
        LC_COLLATE = "C",
        LANG = "en_US.UTF-8"
    are supported and installed on your system.
perl: warning: Falling back to the standard locale ("C").

It's mostly harmless, but may distract from actual errors. I don't know
if this is an issue on other distributions, nor how to fix it there.


kimrhh:
:+1: been annoyed by that for a long time as well :)

ravi-prevas:
Yeah, and centos and opensuse already rely on that as wc -l reveals :). Is there some easy way for me to check whether the issue also exists for the other distros?

kimrhh:
@esben 6.3 material

PR #171, opened 2016-10-05
All the way to parallel tasks, and a little beyond

Villemoes:
This is a complete branch getting to parallel task execution. The first 13 commits are identical to PR#168 (except for a rebase to current master).

The diffstat shows that we exclusively touch files under lib/oelite, and I've doublechecked that this doesn't change meta-data hashes (by doing oe bake world on master, checking out this branch, and seeing that oe bake world says "nothing to do").

There are admittedly some leftovers from the development (such as initially setting the oven capacity to a hardcoded 8, later changing that to something based on $PARALLEL_MAKE), and I'm open to reordering/squashing/etc. But I'm not doing that unless some reviewer finds it important and points out specific problems. AFAICT, the worst issue is the "temporary" disabling of --fake-build, which doesn't get reinstated in this series (and, in fact, I do not yet have a patch for doing that).

The speed-up for an oe bake world has consistently been 40-50% during all my testing, where I've done "rm -rf tmp/stamp/machine tmp/work/machine" between runs (i.e., I haven't rebuilt the toolchain). I suppose those numbers would be slightly smaller if one compares completely clean builds.

2016-10-10: Pushed new version, taking into account comments from mnhu (on #168) and kimrhh. Some profile writing moved from baker.py to oven.py. With --debug we now cat the log file to stdout when the task is done. oven.py implementation hopefully made slightly more descriptive.

### DIFF
--- a/lib/oelite/baker.py
+++ b/lib/oelite/baker.py
@@ -259,6 +259,7 @@ class OEliteBaker:
     def bake(self):
 
         self.setup_tmpdir()
+        oelite.profiling.init(self.config)
 
         # task(s) to do
         if self.options.task:
--- a/lib/oelite/baker.py
+++ b/lib/oelite/baker.py
@@ -11,6 +11,7 @@ import oelite.task
 import oelite.item
 from oelite.parse import *
 from oelite.cookbook import CookBook
+import oelite.profiling
 
 import oelite.fetch
 
@@ -89,6 +90,7 @@ def add_show_parser_options(parser):
 
 class OEliteBaker:
 
+    @oelite.profiling.profile_rusage_delta
     def __init__(self, options, args, config):
         self.options = options
         self.debug = self.options.debug
--- a/lib/oelite/baker.py
+++ b/lib/oelite/baker.py
@@ -569,13 +569,17 @@ class OEliteBaker:
 
         # FIXME: add some kind of statistics, with total_tasks,
         # prebaked_tasks, running_tasks, failed_tasks, done_tasks
-        task = self.runq.get_runabletask()
         rusage = oelite.profiling.Rusage("Build")
         total = self.runq.number_of_tasks_to_build()
         count = 0
         exitcode = 0
+        pending = []
         failed_task_list = []
-        while task:
+        while True:
+            pending += self.runq.get_runabletasks()
+            if not pending:
+                break
+            task = pending.pop()
             count += 1
             debug("")
             debug("Preparing %s"%(task))
@@ -592,7 +596,6 @@ class OEliteBaker:
                 task.build_failed()
                 # FIXME: support command-line option to abort on first
                 # failed task
-            task = self.runq.get_runabletask()
         rusage.end()
 
         if exitcode:
--- a/lib/oelite/baker.py
+++ b/lib/oelite/baker.py
@@ -244,7 +244,7 @@ class OEliteBaker:
             self.runq = OEliteRunQueue(self.config, self.cookbook)
             self.runq._add_recipe(recipe, task)
             task = self.cookbook.get_task(recipe=recipe, name=task)
-            task.prepare(self.runq)
+            task.prepare()
             meta = task.meta()
         else:
             meta = recipe.meta
@@ -579,7 +579,7 @@ class OEliteBaker:
             count += 1
             debug("")
             debug("Preparing %s"%(task))
-            task.prepare(self.runq)
+            task.prepare()
             meta = task.meta()
             info("Running %d / %d %s"%(count, total, task))
             task.build_started()
--- a/lib/oelite/baker.py
+++ b/lib/oelite/baker.py
@@ -593,22 +593,21 @@ class OEliteBaker:
                 self.runq.mark_done(task)
             else:
                 err("%s failed"%(task))
-                exitcode = 1
                 failed_task_list.append(task)
                 task.build_failed()
                 # FIXME: support command-line option to abort on first
                 # failed task
         rusage.end()
 
-        if exitcode:
-             for task in failed_task_list:
-                print "\nERROR: %s failed  %s"%(task,task.logfn)
-                if self.debug_loglines:
-                    with open(task.logfn, 'r') as fin:
-                        if self.debug_loglines < 0:
-                            print fin.read()
-                        else:
-                            print ''.join(fin.readlines()[-self.debug_loglines:])
+        for task in failed_task_list:
+            exitcode = 1
+            print "\nERROR: %s failed  %s"%(task,task.logfn)
+            if self.debug_loglines:
+                with open(task.logfn, 'r') as fin:
+                    if self.debug_loglines < 0:
+                        print fin.read()
+                    else:
+                        print ''.join(fin.readlines()[-self.debug_loglines:])
         return exitcode
 
 
--- a/lib/oelite/baker.py
+++ b/lib/oelite/baker.py
@@ -293,7 +293,7 @@ class OEliteBaker:
         # first, add complete dependency tree, with complete
         # task-to-task and task-to-package/task dependency information
         debug("Building dependency tree")
-        start = oelite.util.now()
+        rusage = oelite.profiling.Rusage("Building dependency tree")
         for task in self.tasks_todo:
             task = oelite.task.task_name(task)
             try:
@@ -311,7 +311,7 @@ class OEliteBaker:
                 die("No such task: %s: %s"%(thing, e.__str__()))
             except oebakery.FatalError, e:
                 die("Failed to add %s:%s to runqueue"%(thing, task))
-        oelite.util.timing_info("Building dependency tree", start)
+        rusage.end()
 
         # Generate recipe dependency graph
         recipes = set([])
@@ -359,7 +359,7 @@ class OEliteBaker:
         task = self.runq.get_metahashable_task()
         total = self.runq.number_of_runq_tasks()
         count = 0
-        start = oelite.util.now()
+        rusage = oelite.profiling.Rusage("Calculating task metadata hashes")
         while task:
             oelite.util.progress_info("Calculating task metadata hashes",
                                       total, count)
@@ -433,7 +433,7 @@ class OEliteBaker:
         oelite.util.progress_info("Calculating task metadata hashes",
                                   total, count)
 
-        oelite.util.timing_info("Calculation task metadata hashes", start)
+        rusage.end()
 
         if count != total:
             print ""
@@ -570,7 +570,7 @@ class OEliteBaker:
         # FIXME: add some kind of statistics, with total_tasks,
         # prebaked_tasks, running_tasks, failed_tasks, done_tasks
         task = self.runq.get_runabletask()
-        start = oelite.util.now()
+        rusage = oelite.profiling.Rusage("Build")
         total = self.runq.number_of_tasks_to_build()
         count = 0
         exitcode = 0
@@ -594,7 +594,7 @@ class OEliteBaker:
                 # FIXME: support command-line option to abort on first
                 # failed task
             task = self.runq.get_runabletask()
-        oelite.util.timing_info("Build", start)
+        rusage.end()
 
         if exitcode:
              for task in failed_task_list:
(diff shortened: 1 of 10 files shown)


lib/oelite/runq.py
```diff
@@ -557,6 +557,11 @@ def get_runabletask(self):
             return None
         return self.cookbook.get_task(id=task_id)
 
+    def get_runabletasks(self):
+        self.update_runabletasks()
+        ret = [self.cookbook.get_task(id=task_id) for task_id in self.runable]
+        self.runable = []
```
kimrhh-prevas: runable will be populated with new runable tasks later on?

lib/oelite/function.py
```diff
@@ -120,6 +121,28 @@ def __init__(self, meta, var, name=None, tmpdir=None):
         super(ShellFunction, self).__init__(meta, var, name, tmpdir)
         return
 
+    def runscript(self, cmd):
```
kimrhh-prevas: runscript? it takes one command or a whole script or?

lib/oelite/task.py
```diff
@@ -183,24 +183,17 @@ def colon_split(s):
...
-        self.logfn = "%s/%s.%s.log"%(function.tmpdir, self.name, meta.get("DATETIME"))
-        self.logsymlink = "%s/%s.log"%(function.tmpdir, self.name)
+        self.cwd = self.do_dirs() or meta.get("B")
+        self.stdin = open("/dev/null", "r")
+        self.logfn = "%s/%s.%s.log"%(self.function.tmpdir, self.name, meta.get("DATETIME"))
+        self.logsymlink = "%s/%s.log"%(self.function.tmpdir, self.name)
         oelite.util.makedirs(os.path.dirname(self.logfn))
         try:
-            if self.debug:
-                logfile = os.popen("tee %s"%self.logfn, "w")
```
kimrhh-prevas: meh, I really do not like to loose that feature, it is highly annoying that you have to dig into logfiles if you are running a specific build to just quickly see what is happening e.g. for configure

However, I'm pretty sure the way the output forwarding was done with tee, introduced a massive slowdown to the build, so that was obviously not nice anyway

It is not necessary to force it into serial mode here to show output. The Exherbo package manager already implements this nicely, but muxing between the output from the tasks been run, showing each of them for $user-configurable-amount-of-time. This works just fine in practice

lib/oelite/task.py
```diff
@@ -209,46 +202,61 @@ def run(self):
...
-        real_stdin = os.dup(sys.stdin.fileno())
-        real_stdout = os.dup(sys.stdout.fileno())
-        real_stderr = os.dup(sys.stderr.fileno())
-        os.dup2(stdin.fileno(), sys.stdin.fileno())
-        os.dup2(logfile.fileno(), sys.stdout.fileno())
-        os.dup2(logfile.fileno(), sys.stderr.fileno())
+    def save_context(self):
+        self.old_stdin = os.dup(sys.stdin.fileno())
+        self.old_stdout = os.dup(sys.stdout.fileno())
+        self.old_stderr = os.dup(sys.stderr.fileno())
```
kimrhh-prevas: I know you did not implement this, but shouldn't most of this code handle OSError exceptions?

lib/oelite/oven.py
```diff
@@ -0,0 +1,111 @@
...
+        self.total = baker.runq.number_of_tasks_to_build()
+        self.count = 0
+
+    def add(self, task):
+        self.capacity -= task.weight
+        self.starttime[task] = oelite.util.now()
+
+    def remove(self, task):
+        now = oelite.util.now()
+        delta = (now - self.starttime[task]).total_seconds()
```
kimrhh-prevas: a remove method returning a time delta seems pretty weird

lib/oelite/oven.py
```diff
@@ -0,0 +1,111 @@
...
+
+        task.start()
+
+    def wait_task(self, poll, task):
+        """Wait for a specific task to finish baking. Returns pair (result,
+        delta), or None in case poll=True and the task is not yet
+        done.
+
+        """
+        assert(task in self.starttime)
```
kimrhh-prevas: huh?

lib/oelite/oven.py
```diff
@@ -0,0 +1,111 @@
...
+            time.sleep(0.1)
+        return None
+
+    def wait_all(self, poll):
+        """Do wait_task once for every task currently in the oven once. With
+        poll=False, this amounts to waiting for every current task to
+        finish.
+
+        """
+        tasks = self.starttime.keys()
```
kimrhh-prevas: this starttime mixed with tasks handling is confusing. time and tasks might be related, but not like this

lib/oelite/function.py
```diff
@@ -128,32 +128,43 @@ def __call__(self):
...
+            return None
+        if ret == 0:
+            self.result = True
+        else:
+            self.result = False
+            print "Error: Command failed: %r: %d"%(self.cmdstr, ret)
+        return self.result
+
+
+    def startscript(self, cmd):
```
kimrhh-prevas: again, script/command?

lib/oelite/oven.py
```diff
@@ -34,6 +34,9 @@ def __init__(self, baker, capacity):
         self.total = baker.runq.number_of_tasks_to_build()
         self.count = 0
 
+    def currently_baking(self):
+        return self.starttime.keys()
```
kimrhh-prevas: it's pretty consistent :) still confusing

lib/oelite/oven.py
```diff
@@ -102,13 +103,20 @@ def wait_any(self, poll):
...
+        i = 0
         while True:
             for t in tasks:
                 result = self.wait_task(True, t)
                 if result is not None:
                     return result
             if poll:
                 break
+            i += 1
+            if i == 4 and self.stdout_isatty:
```
kimrhh-prevas: 4 ? I say 42 :P

lib/oelite/oven.py
```diff
@@ -26,7 +26,13 @@
...
 
 class OEliteOven:
-    def __init__(self, baker, capacity):
+    def __init__(self, baker, capacity=None):
+        if capacity is None:
+            pmake = baker.config.get("PARALLEL_MAKE")
+            if pmake is None or pmake == "":
+                capacity = 3
+            else:
+                capacity = int(pmake.replace("-j", "")) + 2
```
kimrhh-prevas: should that really be hardcoded? why not let if be another configurable

lib/oelite/runq.py
Villemoes: Not my invention. Look at how get_runabletask used to be implemented; self.runable is a sort of cache which gets filled on every call (via a call to get_readytasks, which does some sql stuff), and then a single element is popped and returned. All I do is steal the entire cache.

lib/oelite/function.py
Villemoes: It takes one command, which is the path to the script we've written to the .run file. That command is prepended with LC_ALL=C and possibly other stuff, which means that we still need the shell to interpret the command.

lib/oelite/function.py
```diff
@@ -120,6 +121,28 @@ def __init__(self, meta, var, name=None, tmpdir=None):
...
+            retval = None
+            returncode = subprocess.call(cmd, stdin=sys.stdin, shell=True)
+            if returncode == 0:
+                retval = True
+            else:
+                print "Error: Command failed: %r: %d"%(cmdstr, returncode)
+        except OSError, e:
+            if e.errno == 2:
+                print "Error: Command not found:", cmdname
+            else:
```
Villemoes: Yes, but that's quite deliberate. We just want to return False when the command fails. Please look at util.shcmd where this is copied (and simplified) from. This commit is supposed to be a semantic noop.

lib/oelite/function.py
Villemoes: See above. Previously, runscript made sense because we were still doing everything serially. Now we just want to start it, so I changed the name.

lib/oelite/task.py
Villemoes: Well, I'm open to ideas for how to preserve or reinstate this. I can only think of the brute-force way: when the task is done, cat its log-file to the terminal.

lib/oelite/oven.py
Villemoes: Well, as I said, there's room for tuning. But I'm really not a fan of adding a configurable just for the sake of it - how would a user know what a sensible value is?

lib/oelite/runq.py
Villemoes: ... after refactoring the update from sql part into .update_runabletasks, which is where you'll find self.runable being populated

lib/oelite/task.py
esben: Muxing of output sounds a bit painful IMHO.

lib/oelite/oven.py
esben: Thinking a bit out of the box. It could be nice if there were a single setting for the amount of parallel "threads" that were allowed/used.

If this parallel thingy were set to 8, that could be

- 8 parallel tasks running do_configure or other sequential stuff
- 1 task running "make -j8"
- 1 taks running "make -j4" and 4 tasks running sequential stuff

and so on.

And this setting should default to something based on the number of available cpu cores.

kimrhh:
I have been testing this code locally, trying to make it break but so far without much luck :) I have removed random prebakes and directories in tmp/work to see if we might have some nasty assumptions on presence of directions, or build order or stuff like that. It keeps working, so it is very nice

Besides that, we are using this code in production for PIL now, on top of core-5.0.0, and that also works nicely.

I have reviewed the full PR, and besides the missing output feature I don't have anything major to point out. I also think we can postpone the implementation of debug output to a later PR instead of stalling this further.

lib/oelite/task.py
Villemoes: As I wrote in the updated summary above, I now do the "cat the log-file when the task is done".

sknsean:
I have also been running this code locally and i have seen no issues...
I have merged this on my 2 home projects and one of the projects using musl :+1: 

We have been using this code in the last 2 PIL releases. 

Only a little down side, the output can get a little confusing. But i guess it could be added later

jacobbarsoe:
Same here: nice work. I have been running this at home without issues - yet. 

-Now one wants the frontend like what they have in yocto....

esben:
While I am not going to object to it, I will miss the console "tee" of task out in some situations.  But the choice between that and parallel build is not a hard choice :)

Maybe we could re-introduce it in some way later on.  When running with debug, I could live without parallel build.

lib/oelite/compat.py
```diff
@@ -0,0 +1,81 @@
...
+
+__all__ = ["dup_cloexec", "open_cloexec"]
+
+# uname() -> (sysname, nodename, release, version, machine)
+sysname, _, release, _, machine = os.uname()
+
+values = dict()
+if (sysname, machine) == ("Linux", "x86_64"):
+    # these are supported since at least 2.6.24, so just set them unconditionally
+    values['O_CLOEXEC'] = 02000000
```
Villemoes: I agree, that's the only sane way to write octal constants. But this was just copy-pasted directly from the kernel uapi header. If we ever move to python 3, we don't need to monkey-patch in support for os.O_CLOEXEC.

lib/oelite/compat.py
```diff
@@ -0,0 +1,81 @@
...
+    values['F_DUPFD_CLOEXEC'] = 1024+6
+
+# import bb.utils doesn't work when we try to run this by itself (to
+# test it), so this just serves to show what one could do, provided
+# someone figures out how to make the vercmp_string available.
```
Villemoes: Hm, that might work. But it does add another host dependency.

lib/oelite/baker.py
```diff
@@ -567,44 +573,61 @@ def bake(self):
...
-                task.build_done(self.runq.get_task_buildhash(task))
-                self.runq.mark_done(task)
-            else:
-                err("%s failed"%(task))
-                exitcode = 1
-                failed_task_list.append(task)
-                task.build_failed()
-                # FIXME: support command-line option to abort on first
-                # failed task
-            task = self.runq.get_runabletask()
```
Villemoes: Yes.

PR #177, opened 2016-10-21
a little faster hash computation

Villemoes:
This seems to cut around 12% of hash computation.


sknsean:
I have been running this code for 14 days now, i have seen no issues with it.
Nice to get oe bake a little faster :+1:

PR #178, opened 2016-10-26
url.py: Replace urlgrabber with a simple syscall to wget

sknsean:
The underlying problem is that pycurl in ubuntu and debian based
systems are linked with gnutls (instead of openssl)
This results in rror: gnutls_handshake() failed

Signed-off-by: Sean Nyekjaer sean.nyekjaer@prevas.dk

### DIFF
--- a/lib/oelite/fetch/url.py
+++ b/lib/oelite/fetch/url.py
@@ -2,6 +2,7 @@ import oelite.fetch
 import oelite.util
 import os
 import hashlib
+import subprocess
 from oebakery import die, err, warn, info, debug
 
 class UrlFetcher():
@@ -116,34 +117,26 @@ class UrlFetcher():
 
 
 def grab(url, filename, timeout=120, retry=5, proxy=None, ftpmode=False):
-    import urlgrabber
-    import urlgrabber.progress
-    class SimpleProgress(urlgrabber.progress.BaseMeter):
-        def _do_end(self, amount_read, now=None):
-            print "grabbed %d bytes in %.2f seconds" %(amount_read,self.re.elapsed_time())
-
     print "Grabbing", url
-    def grab_fail_callback(data):
-        # Only print debug here when non fatal retries, debug in other cases
-        # is already printed
-        if (data.exception.errno in retrycodes) and (data.tries != data.retry):
-            print "grabbing retry %d/%d, exception %s"%(
-                data.tries, data.retry, data.exception)
-    try:
-        retrycodes = urlgrabber.grabber.URLGrabberOptions().retrycodes
-        if 12 not in retrycodes:
-            retrycodes.append(12)
-        if not os.path.exists(os.path.dirname(filename)):
-            os.makedirs(os.path.dirname(filename))
-        downloaded_file = urlgrabber.urlgrab(
-            url, filename,timeout=timeout,retry=retry, retrycodes=retrycodes,
-            progress_obj=SimpleProgress(), failure_callback=grab_fail_callback,
-            copy_local=True, proxies=proxy, ftp_disable_epsv=ftpmode)
-        if not downloaded_file:
-            return False
-    except urlgrabber.grabber.URLGrabError as e:
-        warn('URLGrabError %i: %s' % (e.errno, e.strerror))
-        if os.path.exists(filename):
-            os.unlink(filename)
+
+    if not os.path.exists(os.path.dirname(filename)):
+        os.makedirs(os.path.dirname(filename))
+
+    cmd = ['wget', '-t', str(retry), '-T', str(timeout), '--passive-ftp', '--no-check-certificate', '--progress=dot:mega', '-v', url, '-O', filename]
+
+    returncode = subprocess.call(cmd)
+
+    if returncode != 0:
+        err("Error %s %d" % (cmd, returncode))
         return False
+
+    if not os.path.exists(filename):
+        err("The fetch command returned success for url %s but %s doesn't exist?!" % (uri, filename))
+        return False
+
+    if os.path.getsize(filename) == 0:
+        os.remove(filename)
+        err("The fetch of %s resulted in a zero size file?! Deleting and failing since this isn't right." % (uri))
+        return False
+
     return True
--- a/conf/provided/default.conf
+++ b/conf/provided/default.conf
@@ -9,6 +9,5 @@ native:bzip2 \
 native:texinfo native:makeinfo native:doxygen \
 native:texlive-extra-utils native:texlive-latex-extra native:latex-xcolor \
 native:flex native:bison \
-native:wget \
 native:libacl1-dev \
 "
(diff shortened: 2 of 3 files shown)


lib/oelite/fetch/url.py
```diff
@@ -116,34 +117,34 @@ def get_disable_ftp_epsv(self, d):
...
-        def _do_end(self, amount_read, now=None):
-            print "grabbed %d bytes in %.2f seconds" %(amount_read,self.re.elapsed_time())
-
     print "Grabbing", url
     def grab_fail_callback(data):
         # Only print debug here when non fatal retries, debug in other cases
         # is already printed
         if (data.exception.errno in retrycodes) and (data.tries != data.retry):
             print "grabbing retry %d/%d, exception %s"%(
                 data.tries, data.retry, data.exception)
```
Villemoes: This grab_fail_callback function becomes unused, so might as well be removed as well.

lib/oelite/fetch/url.py
```diff
@@ -116,34 +117,34 @@ def get_disable_ftp_epsv(self, d):
...
+
+    if not os.path.exists(os.path.dirname(filename)):
+        os.makedirs(os.path.dirname(filename))
+
+    cmd = ['/usr/bin/env', 'wget', '-t', retry, '-T', timeout, '--passive-ftp', '--no-check-certificate', '--progress=dot', '-v', url, '-O', filename]
+
+    p = subprocess.Popen(cmd, stderr=subprocess.PIPE, stdout=subprocess.PIPE)
+
+    stdout, stderr = p.communicate()
+
```
Villemoes: I think it would be better to use the default values for stderr and stdout (namely None). That way all output from wget goes directly to our log file, instead of having to print that to the log ourselves if something fails. In the two sanity checks below, for example, the stderr, stdout strings are not printed in the err() call, making it much harder to figure out what went wrong.

lib/oelite/fetch/url.py
```diff
@@ -116,34 +117,34 @@ def get_disable_ftp_epsv(self, d):
...
-    except urlgrabber.grabber.URLGrabError as e:
-        warn('URLGrabError %i: %s' % (e.errno, e.strerror))
-        if os.path.exists(filename):
-            os.unlink(filename)
+
+    if not os.path.exists(os.path.dirname(filename)):
+        os.makedirs(os.path.dirname(filename))
+
+    cmd = ['/usr/bin/env', 'wget', '-t', retry, '-T', timeout, '--passive-ftp', '--no-check-certificate', '--progress=dot', '-v', url, '-O', filename]
+
```
Villemoes: Why use /usr/bin/env? subprocess.Popen uses execvp, so wget would be search for in $PATH just as env ends up doing - but if for some reason /usr/bin/env doesn't exist, this will fail.

lib/oelite/fetch/url.py
```diff
@@ -116,34 +117,34 @@ def get_disable_ftp_epsv(self, d):
...
+    stdout, stderr = p.communicate()
+
+    if p.returncode != 0:
+        err("Error %s %d %s %s" % ( cmd, p.returncode, stdout, stderr))
         return False
+
+    if not os.path.exists(filename):
+        err("The fetch command returned success for url %s but %s doesn't exist?!" % (uri, ud.localpath), uri)
+	return False
+
```
Villemoes: There seems to be some whitespace damage here. Mixed tabs/spaces? Also, should ud.localpath be filename? Is the trailing uri argument on purpose?

lib/oelite/fetch/url.py
```diff
@@ -116,34 +117,34 @@ def get_disable_ftp_epsv(self, d):
...
+    if os.path.getsize(filename) == 0:
+        os.remove(filename)
+        err("The fetch of %s resulted in a zero size file?! Deleting and failing since this isn't right." % (uri), uri)
+	return False
+
```
Villemoes: again, whitespace damage and extra argument.

lib/oelite/fetch/url.py
esben: How much output (noise) is wget generating with this change?  I would really like to not see progress information like spinning wheels or anything like that in the log...

Villemoes: I actually think some progress information can be useful. One never looks at the logs on success, but when stuff fails, I like to see as much as possible. Of course the log shouldn't contain terminal control chars or other binary mess, which wget is smart enough not to do when stderr is not a tty. In any case, here's what the above would roughly generate:

`$ wget -v --progress=dot http://ftp.gnu.org/gnu/screen/screen-4.2.1.tar.gz 
--2016-11-01 15:30:11--  http://ftp.gnu.org/gnu/screen/screen-4.2.1.tar.gz
Resolving ftp.gnu.org (ftp.gnu.org)... 208.118.235.20, 2001:4830:134:3::b
Connecting to ftp.gnu.org (ftp.gnu.org)|208.118.235.20|:80... connected.
HTTP request sent, awaiting response... 200 OK
Length: 1062485 (1,0M) [application/x-gzip]
Saving to: ‘screen-4.2.1.tar.gz’

     0K .......... .......... .......... .......... ..........  4%  207K 5s
    50K .......... .......... .......... .......... ..........  9%  418K 3s
   100K .......... .......... .......... .......... .......... 14% 11,4M 2s
   150K .......... .......... .......... .......... .......... 19% 10,9M 2s
   200K .......... .......... .......... .......... .......... 24%  435K 2s
   250K .......... .......... .......... .......... .......... 28% 11,3M 1s
   300K .......... .......... .......... .......... .......... 33% 11,0M 1s
   350K .......... .......... .......... .......... .......... 38% 11,2M 1s
   400K .......... .......... .......... .......... .......... 43%  474K 1s
   450K .......... .......... .......... .......... .......... 48% 11,3M 1s
   500K .......... .......... .......... .......... .......... 53% 11,3M 1s
   550K .......... .......... .......... .......... .......... 57% 11,0M 0s
   600K .......... .......... .......... .......... .......... 62% 11,0M 0s
   650K .......... .......... .......... .......... .......... 67% 10,9M 0s
   700K .......... .......... .......... .......... .......... 72% 11,3M 0s
   750K .......... .......... .......... .......... .......... 77% 11,0M 0s
   800K .......... .......... .......... .......... .......... 81%  577K 0s
   850K .......... .......... .......... .......... .......... 86% 11,3M 0s
   900K .......... .......... .......... .......... .......... 91% 11,0M 0s
   950K .......... .......... .......... .......... .......... 96% 11,3M 0s
  1000K .......... .......... .......... .......              100% 11,3M=0,7s

2016-11-01 15:30:12 (1,37 MB/s) - ‘screen-4.2.1.tar.gz’ saved [1062485/1062485]

`

One could reduce the number of lines by doing `--progress=dot:binary` instead (384K instead of 50K per line), or even more with mega (64K per dot, 3M per line).

esben: dot:mega sounds good.

PR #182, opened 2016-11-04
Some file descriptor tweaks

Villemoes:
The first three are mostly cosmetic, but also adds a little hardening and a very small performance benefit. The fourth is what I believe to be the proper fix for something that was hacked around in oebakery eons ago.

### DIFF
--- a/lib/oelite/task.py
+++ b/lib/oelite/task.py
@@ -198,7 +198,7 @@ class OEliteTask:
         self.function = meta.get_function(self.name)
         self.do_cleandirs()
         self.cwd = self.do_dirs() or meta.get("B")
-        self.stdin = open("/dev/null", "r")
+        self.stdin = open_cloexec("/dev/null", os.O_RDONLY)
         self.logfn = "%s/%s.%s.log"%(self.function.tmpdir, self.name, meta.get("DATETIME"))
         self.logsymlink = "%s/%s.log"%(self.function.tmpdir, self.name)
         oelite.util.makedirs(os.path.dirname(self.logfn))
@@ -216,7 +216,7 @@ class OEliteTask:
         self.saved_stdio = oelite.util.StdioSaver()
 
     def apply_context(self):
-        os.dup2(self.stdin.fileno(), sys.stdin.fileno())
+        os.dup2(self.stdin, sys.stdin.fileno())
         os.dup2(self.logfilefd, sys.stdout.fileno())
         os.dup2(self.logfilefd, sys.stderr.fileno())
 
@@ -224,7 +224,7 @@ class OEliteTask:
         self.saved_stdio.restore(True)
 
     def cleanup_context(self):
-        self.stdin.close()
+        os.close(self.stdin)
         os.close(self.logfilefd)
         if os.path.exists(self.logfn):
             if os.path.getsize(self.logfn) == 0:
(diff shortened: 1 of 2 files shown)


esben:
I am not entirely convinced you are looking at the right problem (as the one that caused the eonic commit in bakery).

When running builds in something archaic as Jenkins CI, stdout and stderr ends up being buffered, and separate, making it very hard to follow what goes on. Why do do you think that unbuffered stderr makes perfect sense?

But with the whole parallel tasks change, things have definitely changed so much with regards to task output, that there might not be a real problem anymore.

But for the "when we can assume that everybody uses a new enough meta/core, ...", I am not sure if/when we will reach that.  I know for certain that I am maintaining several old BSP versions for a long time here, and would appreciate not having to carry around different versions of bakery in my PATH.

Villemoes:
Maybe I've solved another problem, but the problem is definitely real: If stdout is buffered, dup2'ing something over fd 1 will end up sending some output to the wrong file unless one flushes stdout before the dup2. That's very easy to reproduce with the chunk I included (you have to actually read some log files under tmp/work to see it), and I assumed this was the FUBAR referred to, and that the "unbuffer stdout/stderr" simply papered over this.

If you run "command > out.log 2> err.log", I don't see how buffering or non-buffering of those streams make any difference as to correlating the contents of those files. But I assume this is more about some supervisor starting command with fd 1 and 2 referring to separate pipes which the supervisor reads from and then possibly colorizes the stderr output(?) - in that case, yes, buffering up 4096 bytes or whatever of stdout before writing it out would be confusing. I'd probably have done line buffering rather than unbuffering, though, since potentially mixing up stdout and stderr at the character level could be just as confusing...

stderr _is_ unbuffered unless you take very specific action from within the app (see e.g. http://stackoverflow.com/questions/3723795/is-stdout-line-buffered-unbuffered-or-indeterminate-by-default). I don't know if the python runtime does something to change that, but I don't think so. As for why I think it makes perfect sense (and presumably why that's the default in everything unix): you want errors and warnings out as fast as possible instead of having it linger in some internal app buffer, and the only reason for buffering is performance, which is irrelevant for something that's only supposed to be used in exceptional situations.

I actually don't think the parallel task stuff has changed task output - it shouldn't have! Yes, the stuff that's printed to the terminal has changed (replacing "Running foo" with "Foo started" and "Foo completed" plus the occasional "waiting for"), but what goes into task log files from the oe process should be the same.

To summarize: flushing a stream before replacing its associated fd from under it is the right thing to do. If that stream happens to be unbuffered, the flush is a (very cheap) noop, but I'd really like to get rid of an assumption in meta/core about the state of the stdout/stderr streams. Whether we can ever remove the unbuffering hack from oebakery is for later.

esben:
I am not questioning the correctness of this PR.  It is perfectly fine, and I understand the issues with buffering when dup2'ing.

My comment was not directed at the actual changes in this PR, but rather the comment about how this could allow us to drop the stdout/stderr hackery in bakery.

AFAIR, there were also some nuisance about stdout/stderr buffering, where Jenkins output was not really very usable with stdout and stderr being buffered.  But as task output is now sent to a file, and only printed to output on failure, this problem is not relevant with respect to task output.

But bake output, such as "foo started" would be nice to have line buffered, even if being sent to some stupid CI which does not provide a proper tty context for the output.

But to make it short, this whole discussion belongs to bakery, so let's postpone it until you make a PR on that :)

PR #183, opened 2016-11-04
Two leftover oven patches

Villemoes:
I left these out of the big parallel PR mostly because I had to draw a line somewhere.

### DIFF
--- a/lib/oelite/task.py
+++ b/lib/oelite/task.py
@@ -243,12 +243,15 @@ class OEliteTask:
         self.apply_context()
 
         try:
+            prefunc_start = oelite.util.now()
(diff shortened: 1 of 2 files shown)


lib/oelite/task.py
```diff
@@ -274,6 +277,7 @@ def wait(self, poll=False):
         if self.result is not None:
             # Something bad happened in start
             self.cleanup_context()
+            self.func_end = oelite.util.now()
```
Villemoes: Good catch, probably a refactoring artifact. Moreover, there's no reason to store func_end in self, we can just use a local variable (like prefunc_start). Will push a fixed version.

PR #184, opened 2016-11-04
more profiling

Villemoes:
These patches don't really do anything, except add more profiling infrastructure and uses of the existing. These have formed the base of some the recent performance PRs, but I'm pretty sure there's still more to gain, so I'd like to have these upstream.

### DIFF
--- a/lib/oelite/cookbook.py
+++ b/lib/oelite/cookbook.py
@@ -33,7 +33,7 @@ class CookBook(Mapping):
         if not self.db:
             raise Exception("could not create in-memory sqlite db")
         self.db.text_factory = str
-        self.dbc = self.db.cursor()
+        self.dbc = CursorWrapper(self.db.cursor(), profile=False)
         self.init_db()
         self.recipes = {}
         self.packages = {}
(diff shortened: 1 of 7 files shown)


lib/oelite/cookbook.py
```diff
@@ -33,7 +33,7 @@ def __init__(self, baker):
         if not self.db:
             raise Exception("could not create in-memory sqlite db")
         self.db.text_factory = str
-        self.dbc = self.db.cursor()
+        self.dbc = CursorWrapper(self.db.cursor(), profile=False)
```
Villemoes: yeah. I plan on introducing __profile,  __profile_sql, __profile_foo variables one can set in local.conf to control whether and which profiling data gets gathered, but just haven't got around to it yet. But being able to turn this on by changing just a single line is still a lot easier than bolting on the rest of the infrastructure.

esben: :+1:

lib/oelite/dbutil.py
```diff
@@ -1,4 +1,164 @@
...
+            return self
+        return ret
+
+    def __iter__(self):
+        return self
+
+    # This is Python 2, so we have to define .next rather than .__next__ .
+    def next(self):
+        if not self.profile:
+            # This should never really happen.
```
esben: Ok.

PR #186, opened 2016-11-09
oelite/meta: faster deep-copying

Villemoes:
The deepcopy function in Python's standard library is extremely
slow. I've hacked together a C implementation of parts of copy.deepcopy,
which I'm in the process of upstreaming to CPython - but it will never
get applied/backported to 2.7, so here's a version we can use.

This cuts 60% from the deepcopying time, which translates to 25% of the
metadata hash computation time. I'm somewhat surprised the gain is not
larger than that, but we do have a few somewhat large/deep structures
hidden in e.g. the __overrides and __mtimes "variables".

For now, it does require an explicit action from the user, and having
Python headers available: Go to lib/oelite/meta and execute

  gcc -g -O2 -o _copy.so -fPIC -shared -I/usr/include/python2.7 _copy.c

I don't know if there's a good way of doing this automatically or
advising the user to do it - the try..except does ensure that there's no
ill effect of not reading meta/core commit logs carefully.


esben:
The good:

1. It is nice to improve things :)
2. Everything works fine without user doing anything special.

The bad:

1. When information on how to take advantage of this is added to the OE-lite documentation (which it obviously should), we are clearly going in the wrong direction with regards to KISS.

Villemoes:
Longer term we could add the C file to oebakery and have it built by the distutils/setuptools/whateverthecrapiscallednowadays. This just makes it available now. To merge or not to merge is your call, obviously - it's small and unintrusive enough that anyone who might want it anyway could apply it locally.

Villemoes:
closing, will make a different pull request compiling this automagically.

PR #189, opened 2016-11-10
binutils: add version 2.27

sknsean:


esben:
@sknsean Will you take a look at what might be going on here?

PR #190, opened 2016-11-11
Prepare for saner use flag handling

Villemoes:
These patches are all currently semantic noops, but they serve a few purposes: the prefunc/postfunc patches try to standardize on one idiom for running a pre/postfunc dependent on a use flag, and in doing so actually avoids running them if the use flag is not set (as opposed to doing an early return). Other patches change initializations of DEFAULT_USE_foo from using "0"/"1" to using False/True, which the lexer currently treats exactly equivalently (namely, storing the "0"/"1" string as the value), but they serve to document which USE flags are really meant as boolean flags, as opposed to those where the string value is actually used for something.

All patches are also preparation for potential future changes of the USE flag handling and the interpretation of the False/True tokens, but regardless of what thinks of such changes, I believe these patches are worthwhile in themselves.

Background (added 2016-11-17): USE flags are used and abused for different purposes, and their semantics in current OE-lite sometimes have some surprising consequences. Roughly, USE flags can be divided in two groups (but there may be some that fall outside of these categories, as well as some that straddle them): those that work entirely as boolean flags and are used as the conditionals in override assignments/appends/prepends (e.g. USE_sqlite3_readline), and those whose string value is used for some purpose (e.g. USE_rootfs_ext2_options). 

Now, in a given recipe which has foo in either RECIPE_FLAGS or CLASS_FLAGS, USE_foo gets its value from the first defined of RECIPE_USE_foo, LOCAL_USE_foo, MACHINE_USE_foo, DISTRO_USE_foo and finally DEFAULT_USE_foo. If none of these exist, the variable USE_foo also doesn't exist, and override assignments controlled by USE_foo do not get applied. The same happens if the value it would have is either the empty string or the string "0".

For boolean flags this is fine, and the behaviour is generally as expected from setting such USE flags to "1" or "0", since the values of boolean USE flags are never used directly, so their existence or non-existence by themselves is irrelevant.

HOWEVER, this means, for example, that it is not possible to set MACHINE_USE_rootfs_ext2_options = "", since this would cause a parse error when the (non-existing) variable USE_rootfs_ext2_options is expanded. Due to the way that particular variable is used, one could hack around this by assigning a string consisting of a few spaces, but that's obviously quite ugly, and similar workarounds may not always be applicable. Similar problems occur if one tries to set a configurable number and one just happens to wish it to have the value 0 (I'm looking at you, USE_busybox_feature_beep_length_ms).

This USE flag vanishing is rather confusing, and I aim to change it in some future meta/core release. For now, there is a lot we can do to make the world saner without changing any semantics, but which will prepare for such a change. It is a little known/used fact that the OE-lite language actually allows the tokens True and False on the right-hand side of an assignment. The lexer (currently) translates these to the strings "1" and "0", respectively. This means that we can change all initializations of boolean type USE flags (that is, the {DEFAULT,DISTRO,...}_USE_foo = "0"/"1") to use those tokens, which will help document what foo is (supposed to be) used for, without affecting semantics. Everybody is strongly encouraged to update their own recipes/distro/machine configs similarly.

Some string-type USE flags also need a sentinel value to mean "disable". An example of this is the USE_foo_sysvinit_stop, which historically has been set to "0" to omit the stop script being installed. In this case it is also advised to change to False (which, again, can be done without changing semantics) to highlight the difference between eliding the stop script ("0") vs. installing it at priority level "00".

The biggest problem with unconditionally defining USE variables to whatever value they have according to the RECIPE/LOCAL/... hierarchy comes from python code that does something like 'if d.get("USE_foo"): do something', since, in Python, the string "0" is true while None (the return value from get() when USE_foo doesn't exist) is False. I believe such python code is mostly concentrated in meta/core (we fix/prepare some of it in this series), and it should certainly be rare in random recipes found in the wild, but it is nevertheless such a large change that it should only happen with a couple of release's notice.

### DIFF
--- a/classes/sysvinit.oeclass
+++ b/classes/sysvinit.oeclass
@@ -7,7 +7,13 @@
 ## <chosen name> (declare this to RECIPE_FLAGS)
 ## SYSVINIT_SCRIPT_<chosen name> Set this to the name of your initscript.
 ## <chosen name>_sysvinit_start Set this to the start level.
-## <chosen name>_sysvinit_stop (optional) Set this to the stop level.
+## <chosen name>_sysvinit_stop Set this to the stop level.
+##
+## The level must be a two-digit string. To omit the start/stop
+## script, set <chosen name>_sysvinit_start / <chosen
+## name>_sysvinit_stop to False, e.g.
+##
+##   DEFAULT_USE_sshd_sysvinit_stop = False
 ##
 ## @useflag sysvinit Enable or disable sysvinit. Enabled by default by base
 ##          distro conf.
@@ -28,6 +34,7 @@ CLASS_FLAGS += "sysvinit"
 do_install[postfuncs] += "do_install_sysvinit"
 python do_install_sysvinit () {
     import os
+    import re
 
     path = d.get("D")
     os.chdir(path)
@@ -49,6 +56,17 @@ python do_install_sysvinit () {
         prio = d.get("USE_" + option)
         if not prio:
             continue
+        # "0" is the legacy way of disabling, and also what the token
+        # False currently translates to.
+        if prio == "0":
+            continue
+
+        # Anything but a two-digit string (or False, or "0") is a hard
+        # error.
+        if not re.match("[0-9][0-9]$", prio):
+            bb.error("Invalid value of 'USE_%s': %s" % (option, prio))
+            bb.error("Must be two-digit string or False")
+            return False
 
         if start_symlink:
             name = option[0:-len("_sysvinit_start")]
--- a/classes/sysvinit.oeclass
+++ b/classes/sysvinit.oeclass
@@ -31,15 +31,15 @@ RDEPENDS_${PN}:>USE_sysvinit = " ${RDEPENDS_SYSVINIT}"
 
 CLASS_FLAGS += "sysvinit"
 
-do_install[postfuncs] += "do_install_sysvinit"
+DO_INSTALL_SYSVINIT = ""
+DO_INSTALL_SYSVINIT:USE_sysvinit = "do_install_sysvinit"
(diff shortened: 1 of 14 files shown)


classes/sysvinit.oeclass
```diff
@@ -7,7 +7,11 @@
 ## <chosen name> (declare this to RECIPE_FLAGS)
 ## SYSVINIT_SCRIPT_<chosen name> Set this to the name of your initscript.
 ## <chosen name>_sysvinit_start Set this to the start level.
-## <chosen name>_sysvinit_stop (optional) Set this to the stop level.
+## <chosen name>_sysvinit_stop Set this to the stop level.
+##
+## To omit the start/stop script, set <chosen name>_sysvinit_start /
+## <chosen name>_sysvinit_stop to anything but a two-digit string, the
+## string "disable" being the preferred way.
```
esben: It would be really nice to be able to assign None instead...

esben:
Should we merge this for 6.3, or postpone it to 7.0?  It should be pretty innocent...

classes/sysvinit.oeclass
Villemoes: I think False would convey the same intention, and that would allow us to replace a lot of existing instances, and still be backwards-compatible with code that uses "0", and forward-compatible with a potential semantic change of the False/True tokens.

Villemoes:
IMO, the sooner we merge this and make some mention of it in the release notes, the sooner downstreams can learn to update their distros and machine configs, and the sooner can we make the world saner. But first let me know what you think of using False for the sysvinit_stop thing.

esben:
I am okay with using False.  Much better than current situation, although syntactically I would have preferred None (but we do not have lexer/parser support for that, so I am not really pushing that).

esben:
Could you wrap up some text for the release notes about this?

PR #193, opened 2016-11-16
classes/kernel: add useflag kernel_dts_vendor

sknsean:


classes/kernel.oeclass
```diff
@@ -287,12 +291,12 @@ do_compile_kernel_dtc() {
 
 KERNEL_COMPILE_POSTFUNCS:>USE_kernel_dts += " do_compile_kernel_dts"
 do_compile_kernel_dts() {
-	oe_runmake "${KERNEL_DTB_FILE}"
+	oe_runmake "${KERNEL_DTB_VENDOR}/${KERNEL_DTB_FILE}"
```
mnhu: I was afraid the this would break if `${KERNEL_DTB_VENDOR}` is empty, but my test shows otherwise:
```
% make ARCH=arm CROSS_COMPILE=arm-cortexa9neon-linux-gnueabi- /imx6dl-wandboard.dtb
  DTC     arch/arm/boot/dts//imx6dl-wandboard.dtb
```

mnhu: I don't really mind the heading slash...

classes/kernel.oeclass
```diff
@@ -248,18 +249,20 @@ UIMAGE_KERNEL_OUTPUT = ""
...
 KERNEL_DTS_DIR:USE_kernel_dts_dir = "${USE_kernel_dts_dir}"
 KERNEL_DTS_FILE = "${MACHINE}.dts"
 KERNEL_DTS_NAME = "${MACHINE}"
+KERNEL_DTB_VENDOR = ""
 KERNEL_DTB_FILE = ""
 
 addhook kernel_devicetree_init to post_recipe_parse after set_useflags
 def kernel_devicetree_init(d):
     kernel_dtb = d.get('USE_kernel_external_dtb')
+    kernel_dts_vendor = d.get('USE_kernel_dts_vendor')
```
esben: Without any setting of USE_kernel_dts_vendor, will we end up with None?  And how is that handled later on?

sknsean: does the @mnhu comments answer this question?

esben: Ok, I will re-formulate.  Have you tested this change with USE_kernel_dts set, but without USE_kernel_dts_vendor set?

sknsean: Yes and it fails :-1: 
`CRITICAL: bake failed: Exception: Cannot expand variable ${KERNEL_DTB_VENDOR`

sknsean: Now have tested this with and without USE_kernel_dts_vendor set...

mnhu: Better do `kernel_dts_vendor = d.get('USE_kernel_dts_vendor') or ''` instead of the conditional below.

PR #194, opened 2016-11-18
remove lots of unused code

Villemoes:
I think we can safely remove all of this. I'd like to get rid of the rest of lib/bb at some point, but this is the easy part.


lib/oelite/log.py
```diff
@@ -1,22 +0,0 @@
-import logging
```
esben: Hmm... What do you think about the idea behind this module?  Maybe we should start cleaning up debug and logging instead of just removing the only (although completely stalled) attempt at doing something about it.

kimrhh: while I really miss a proper logging implementation in OE-lite, it is probably better to start by ripping away all the unused parts, before starting to implement new stuff. non-working features like this one adds to the confusion when reading through the code

It would be really nice to talk about how we could implement proper logging though instead of having to insert new print statements all-over the place every time we need to debug things :(

Villemoes: I like the idea of an oelite.log module, and I've done some thinking about how to implement it. But it won't be implemented in terms of the python stdlib logging module, since it's way too hard to figure out what happens inside that. Instead, since porting would have to happen gradually and we'd have to stay compatible with bare print statements for a long while, it would be built from scratch, writing to sys.stdout and/or sys.stderr, be made "context-aware" (switching file descriptors to a task log file and back to the terminal would be handled in oelite.log module); that way one could for example easily keep an fd for the terminal open and make oelite.log.<configurable bad enough level> made in "task" context also be echoed immediately to the terminal.

The initial implementation would probably just make all oelite.log.foo aliases for print (+ the fd handling), and then one can gradually change the codebase over.

So I agree with kimrhh, it's easier to start from scratch when we get around to doing something about this, which it seems we all want to see happening.

esben: Ok, sounds like a fine plan.

PR #199, opened 2016-12-05
optimize libmagic use

Villemoes:
runstrip opens, parses and closes the system's "magic" database for every file it processes, which is somewhat wasteful.

Each do_chrpath leaks the open database handle, which amounts to 700MB+ of extra VM during oe bake world.

Since we only ever use the default database with default flags, just open and parse the database once.

### DIFF
--- a/lib/oelite/magiccache.py
+++ b/lib/oelite/magiccache.py
@@ -1,0 +1,41 @@
+import magic
+
+class OEliteMagic:
+
+    def __init__(self, flags, filename):
+        self.refs = 1
+        self.magic  = magic.open(flags)
+        self.filename = filename
+        self.magic.load(self.filename)
+
+    def get(self):
+        self.refs += 1
+        return self
+
+    def file(self, filename):
+        return self.magic.file(filename)
+
+    def descriptor(self, fd):
+        return self.magic.descriptor(fd)
+
+    def close(self):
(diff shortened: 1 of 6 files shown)


lib/oelite/magiccache.py
```diff
@@ -0,0 +1,46 @@
...
+            self.magic = None
+
+
+cache = dict()
+
+def open(flags = magic.MAGIC_NONE, filename=None):
+    t = (flags,filename)
+    if t not in cache:
+        cache[t] = OEliteMagic(flags, filename)
+    return cache[t].get()
```
esben: This will set `refs=2` for first open. Is that intended?

Villemoes: Yes. The cache owns one reference (the initial one created by the constructor), the caller obtains another. When the caller is done, he calls close(), which decrements refs, but the object is kept alive precisely by the reference from the cache dict. If somebody calls clear_cache(), the cache dict drops all its references, but any OEliteMagic in active use is kept alive by the reference that was handed out. 

This is of course all very academic since we're only ever going to have one element in the cache, and we're never calling clear_cache (I've done it for debugging), but it felt cleaner to do it this way rather than hardcoding the singleton.

esben: Ah, I see :)

PR #200, opened 2016-12-06
fetch/url.py fixes and hardening

Villemoes:
Mostly an RFC, but combining this with the async python branch, this survives doing an 'oe bake universe -t fetch' with initially empty ingredients, except for the stuff that would fail anyway (e.g., currently the libwebsockets git repo is unreachable, and git repos are anyway unaffected by this patch).

### DIFF
--- a/lib/oelite/fetch/url.py
+++ b/lib/oelite/fetch/url.py
@@ -4,6 +4,8 @@ import os
 import hashlib
 import subprocess
 from oebakery import die, err, warn, info, debug
+import tempfile
+import errno
 
 class UrlFetcher():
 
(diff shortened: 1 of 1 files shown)


lib/oelite/fetch/url.py
```diff
@@ -84,59 +86,95 @@ def localsignature(self):
...
+    d = os.path.dirname(filename)
+    f = os.path.basename(filename)
+    if not os.path.exists(d):
+        os.makedirs(d)
+
+    # Using mktemp here is perfectly fine - wget with --no-clobber and
+    # -O refuses to write to an existing file. We must download to the
+    # actual ingredient dir rather than e.g. /tmp to ensure that we
+    # can do a link(2) call without encountering EXDEV.
+    dl_tgt = tempfile.mktemp(prefix = f + ".", dir = d)
```
esben: Why not use mkstemp instead, and overwrite the empty temporary file, and finish up with link(2)?

Villemoes: Because in the 1 in 56 billion (that's 62^6, from reading tempfile.py source code) case where two racing processes end up using the same name, I don't want one wget process to silently truncate the output of another. I'd much rather the build fails, which means that it must be wget which does the O_EXCL open. We can reduce the probability further by throwing in getpid(), but PID namespaces (and network file systems) implies that doesn't make it 0.

Villemoes: Hm, wait. As long as wget doesn't unlink the destination and create it itself, mkstemp should actually be fine. I obviously need to think this through a little more.

Villemoes: Ah, I remember now. I read this in wget's man page, which made me decide that we needed -nc:
```
When running Wget without -N, -nc, -r, or -p, downloading the same file in the same directory will result in the original copy of file being preserved and the second copy being named file.1.
```
but maybe we can just make wget write to stdout and give it the fd which we open.

PR #201, opened 2016-12-16
kernel headers update

Villemoes:


recipes/linux-headers/linux-headers.inc
```diff
@@ -4,7 +4,7 @@ inherit kernel-headers
 RECIPE_TYPES = "machine sdk"
 
 require conf/fetch/kernelorg.conf
-SRC_URI = "${KERNELORG_MIRROR}/linux/kernel/v${KERNEL_VERSION_PATCHLEVEL}/linux-${PV}.tar.bz2"
+SRC_URI = "${KERNELORG_MIRROR}/linux/kernel/v${@'${PV}'.split('.')[0]+'.x'}/linux-${PV}.tar.xz"
```
esben: Ok, but correct me if I am wrong.  We are breaking existing pre 3.0 linux recipes with this?  I am not objecting, as the fix is clearly trivial, but it needs to go in the release notes for 7.0.

Villemoes: Well, yes, but is there any way a recipe from outside recipes/linux-headers/ would ever end up including that .inc file? If not, to be affected it would require someone to actively patch in a linux-headers_2.6.oe on top of upstream core releases, which is of course possible, but unlikely. It would be easier to understand your concern if this change happened in a class file.

Villemoes: More to the point: It doesn't make sense to have an .inc file which is unusable by the current in-tree recipes.

esben: Totally agree.  I bluntly missed the context (.inc file and not class).  As far as I am concerned, .inc files are for in-tree use only.

PR #203, opened 2016-12-20
oelite/meta/dict.py: put our dicts on a diet

Villemoes:
It's almost January, so time for the annual diet. Let's include the
dicts backing our metadata instances. This cuts 46% of the memory use of
the .smpl and .cplx members themselves (not accounting for sizes of keys
and values), corresponding to about 10% of the total memory consumption.
For an "oe bake world", that translates to around 63 MB of memory with
the metalayers I have. It doesn't add any noticable time to the
initialization phase.

I admit it's a little hacky, but I think the code-to-comment ratio and
the gain should make up for that. It's a shame Python doesn't expose an
official knob to do this kind of thing.

### DIFF
--- a/lib/oelite/dicttrim.py
+++ b/lib/oelite/dicttrim.py
@@ -1,0 +1,78 @@
+import sys
+
+# Python's dicts never decreases their allocated capacity, even if one
+# deletes most of their members. This means the dicts backing the
+# DictMeta end up being quite oversized after we call filter_meta (in
+# many cases, their fill factor is below 15%). The only way to get rid
+# of that is to recreate the dict from scratch.
+#
+# Now, experimentation and reading CPython source code shows that
+# d.copy() sizes the new dict so it has a fill factor between 25% and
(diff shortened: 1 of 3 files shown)


lib/oelite/dicttrim.py
```diff
@@ -0,0 +1,78 @@
...
+        # reason to set the bar as low as 2/3.
+        return final_fill > 0.85
+
+def trim(d):
+    if len(d) == 0:
+        tmp = {}
+    elif _use_dict_copy(d):
+        tmp = d.copy()
+    else:
+        tmp = dict.fromkeys(dict.fromkeys(d))
```
esben: Is this the extra trick to avoid oversizing with deleted dict entries?

Villemoes: Yes. Might deserve an extra comment, e.g. just appending "(hence the double dict.fromkeys() calls below)" to the last paragraph.

PR #207, opened 2017-01-05
lib/oelite: alternate universe implementation

Villemoes:
The way "oe bake world" and "oe bake universe" is currently implemented
means that the buildbot spends a lot of time building stuff that is
already known to work.

The problem is (roughly) that when a recipe is a "primary" target,
whether to build it again relies on whether the current do_build stamp
file matches the hash we compute for the do_build task, and the files in
the stamp directory reflect (more or less) the status of $WORKDIR. So
when a PR on meta/core updates gcc [1] the buildbot does a complete
rebuild of everything, which is of course what we want.

However, when another PR comes in which just corrects a wrong LICENSE in
some meta/base recipe, we (of course) use the master branch from
meta/core. Since all the do_build stamp files don't match, we build
everything once again, despite the fact that $PACKAGE_DEPLOY_DIR in all
likelyhood contains tar-balls with the appropriate hash values from the
second-to-last build.

So, instead of making every (compatible) recipe a primary target, let us
instead create a phony recipe on the fly which DEPENDS on all the real
recipes. That way, both stamp files as well as tar-balls in
PACKAGE_DEPLOY_DIR are taken into account for deciding what needs
building.

Aside from drastically reducing the amount of work done, this also seems
to make the initialization much faster. For me, the "building dependency
tree" time drops from ~26 seconds to ~5 seconds. We also save 6-8% of
time/memory on metadata hashing, since we don't compute the metadata for
the pseudo-task "build" for all ~300 recipes (only up to and including
do_package).

We keep "universe" as a special token to mean [t + ":world" for t in
recipe_types].

### DIFF
--- a/lib/oelite/cookbook.py
+++ b/lib/oelite/cookbook.py
@@ -78,6 +78,8 @@ class CookBook(Mapping):
         if fail:
             die("Errors while adding recipes to cookbook")
 
+        self.create_world_recipes()
+
         #print "when instantiating from a parsed oefile, do some 'finalizing', ie. collapsing of overrides and append, and remember to save expand_cache also"
 
         return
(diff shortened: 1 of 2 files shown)


Villemoes:
This depends on #208 - I wasn't aware that depending on multiple versions of the same thing wouldn't Just Work(tm).

kimrhh:
I would really really really like to see this merged soon, as it would lower our waste of cpu cycles significantly

lib/oelite/cookbook.py
```diff
@@ -851,3 +853,43 @@ def compute_recipe_build_priorities(self):
             c.remaining_tasks = len(c.tasks)
             del descendants[c]
         assert(len(descendants) == 0)
+
+    def create_world_recipes(self):
+        tmpdir = self.config.get("TMPDIR") or "tmp"
+        recipe_file = os.path.join(tmpdir, "world.oe")
```
esben: could we place them in ${TMPDIR}/recipes instead?

Villemoes: sure, but you do realize there's only one file involved? The plural `recipes` is because that file declares multiple recipe types, so it ends up defining multiple OEliteRecipes.

esben: Yes, I see that.

PR #210, opened 2017-01-12
ncurses pkgconfig/binconfig fixups

Villemoes:
I've hit a few problems related to an autoconf script picking up an ncurses[w]-config script from the host. These fix those issues, and I hope (and believe) they won't cause pain elsewhere.

### DIFF
--- a/recipes/ncurses/ncurses.inc
+++ b/recipes/ncurses/ncurses.inc
@@ -19,7 +19,7 @@ SRC_URI = "${GNU_MIRROR}/ncurses/ncurses-${PV}.tar.gz"
 DEPENDS = "native:libncurses libm"
 DEPENDS:native = "libm"
 
-inherit autotools make-vpath c++
+inherit autotools make-vpath c++ binconfig
 
 export LTCC="${CC}"
 export LTCFLAGS="${CFLAGS}"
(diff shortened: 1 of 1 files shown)


recipes/ncurses/ncurses.inc
```diff
@@ -37,6 +37,8 @@ EXTRA_OECONF += "\
 --without-cxx-binding \
 --with-terminfo-dirs=${sysconfdir}/terminfo:${datadir}/terminfo \
 --enable-overwrite \
+--enable-pc-files \
+--with-pkg-config-libdir=${libdir}/pkgconfig \
```
esben: Why not inherit pkgconfig.oeclass instead?  Is ncurses special in this regard, or do you think we need general improvements to pkgconfig.oeclass?

kimrhh: it only takes a short look at the ncurses configure script to verify this is a special-kind-of-child upstream :) they seem to think it is better to implement everything with their own configure handles

so yes, ncurses is special here

Villemoes: Correct me if I'm wrong, but it seems pkgconfig.oeclass is for setting up appropriate env variables and ensuring a pkg-config binary gets staged. This patch is not about using .pc files from the stuff ncurses depends on; it is about convincing ncurses' build system to provide .pc files for others (those that depend on ncurses) to use.

```
  --with-pkg-config-libdir=XXX use given directory for installing pc-files
  --enable-pc-files       generate and install .pc files for pkg-config
```

esben: Check, and check :)

Issue #74, opened 2015-11-20
try to fix rmwork


kimrhh:
rmwork has been disabled for a long time now, originally since it had bugs in corner cases.

Time to start fixing those

@sknsean enabled it here:

https://github.com/oe-lite/core/compare/master...sknsean:dev-rmwork#diff-c3895395a56eb9a9f1a54295e95046f9R521

help test it, and report back here

sknsean:
1. first problem i ran into was some minor python errors
2. when we are packaging a rootfs or another filesystem, the packages gets deleted before they are used to deploy into tmp/images (check to ugly fix in the link above)

esben:
Sounds strange with the rootfs problem.  Hopefully, rmwork does not delete the packages (as in the package tarballs).  And hopefully, rootfs (or any other RSTAGE based recipe), does not care about the packages in other locations than the tarballs.  It would be helpful to have a log of the discovered problem, and not just a "fix".

And as for the fix, I have a hard time seeing that merged ;)

PR #213, opened 2017-01-20
make rmwork work

Villemoes:
I think I know why rmwork causes problems with recipes that have a do_deploy task (i.e. those that inherit image.oeclass). This reimplementation tries to avoid task ordering issues by simply not doing rmwork as a task, but instead doing it when we know that no more tasks belonging to that recipe remain. Hopefully this can close issue #74 .

For world builds, it is mostly useful when combined with #207 : There is logic in place that disables rmwork for "primary targets", which is everything in the current world implementation.


conf/rmwork.conf
```diff
@@ -1,2 +1,18 @@
...
+${STAGE_UNPACKDIR}
+${RSTAGE_DIR}
+${FSTAGE_DIR}
+${SRCDIR}
+${S}
+${B}
+${D}
+${PKGD}
+${IMAGE_DIR}
+"""
```
esben: Do we really need to list these dirs?  Why not just delete all subdirs of WORKDIR, except tmp?

Villemoes: Well, I think that's what's done up until the last patch, but I thought it might be nice to be able to control it. If one would only use it for saying `RMWORK_DIRS = ""` to disable it for a specific recipe, there's of course a simpler way to do that, but this didn't seem to be much more work. I have no strong feelings either way.

esben: I am fine with keeping it.

PR #215, opened 2017-01-26
oelite/fetch/url.py: fix permissions of generated files

Villemoes:
Downloaded files used to be created with mode 0o644 (modified by umask
as usual), but since we now let tempfile.mkstemp do the open(O_CREAT),
we get 0o600. This is a problem at least for mirroring where oe-lite
runs as one user but the webserver runs as another. The simplest way to
fix it up seems to be an explicit fchmod.

### DIFF
--- a/lib/oelite/fetch/url.py
+++ b/lib/oelite/fetch/url.py
@@ -127,6 +127,12 @@ def grab(url, filename, timeout=120, retry=5, proxies=None, passive_ftp=True):
     # actual ingredient dir rather than e.g. /tmp to ensure that we
     # can do a link(2) call without encountering EXDEV.
     (fd, dl_tgt) = tempfile.mkstemp(prefix = f + ".", dir = d)
+    # Unfortunately, mkstemp() uses mode 0o600 when opening the file,
+    # but we'd rather have used 0o644. So we get to do a little syscall
+    # dance, yay.
+    mask = os.umask(0o022)
+    os.fchmod(fd, 0o644 & ~mask)
+    os.umask(mask)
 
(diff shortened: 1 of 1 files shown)


lib/oelite/fetch/url.py
```diff
@@ -127,6 +127,12 @@ def grab(url, filename, timeout=120, retry=5, proxies=None, passive_ftp=True):
     # actual ingredient dir rather than e.g. /tmp to ensure that we
     # can do a link(2) call without encountering EXDEV.
     (fd, dl_tgt) = tempfile.mkstemp(prefix = f + ".", dir = d)
+    # Unfortunately, mkstemp() uses mode 0o600 when opening the file,
+    # but we'd rather have used 0o644. So we get to do a little syscall
+    # dance, yay.
+    mask = os.umask(0o022)
+    os.fchmod(fd, 0o644 & ~mask)
+    os.umask(mask)
```
esben: Why this umask dance?  Why not just simply set the mode directly?  As far as I can see, os.fchmod/os.chmod does not respect umask anyway.
What obvious detail am I missing?

Villemoes: Exactly because fchmod ignores umask, I have to take it into account manually. If the user has a umask of 0o077, we shouldn't make the files group+other-readable. I want to emulate what the permissions would have been if we'd done open(tmpfile, O_CREAT|O_EXCL, 0o644). And unfortunately one cannot get the current umask without setting a potentially different one as a side effect - the 0o022 is a dummy value which isn't really used for anything, since we undo the setting in line 3.

esben: Makes sense.

PR #219, opened 2017-02-03
metadata: delete various stuff that will never be used

Villemoes:
These patches reduce our memory consumption somewhat as well as make metadata hashing faster. The biggest one is the first which seems to cut 18% of the memory (obviously depending on use flags, architecture etc.); the others are icing I stumbled on a while ago.

### DIFF
--- a/lib/oelite/cookbook.py
+++ b/lib/oelite/cookbook.py
@@ -569,8 +569,9 @@ class CookBook(Mapping):
                 os.remove(cachefile)
 
         for recipe_type in recipes:
-            oelite.pyexec.exechooks(recipes[recipe_type].meta,
-                                    "pre_cookbook")
+            meta = recipes[recipe_type].meta
+            oelite.pyexec.exechooks(meta, "pre_cookbook")
+            meta.trim_unused_overrides()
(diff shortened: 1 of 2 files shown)


esben:
While I tend to agree with you that this most likely not cause any real problems, I cannot 100% convince myself that we are not introducing a slight change in metadata semantics. At least any (currently non-existing) pre_cookbook hooks will not be able to alter overrides anymore.

But again, it does look pretty harmless...

Villemoes:
Well, I suppose one could just do the override trimming after pre_cookbook - without any examples of pre_cookbook hooks, I didn't really know what kind of stuff was supposed to happen there.

esben:
I guess this whole trimming thing is a good example of something that could happen in a pre_cookbook hook ;)

Villemoes:
Yes, except for your excellent point about some other hook function modifying OVERRIDES. This really needs to run after stuff in .oe or .oeclass files can no longer affect the metadata. (OK, sneaky guys could probably put arbitrary stuff in some inline-eval-code-with-side-effects and have it run while we compute hashes, but then they get to keep both pieces...)

PR #223, opened 2017-03-15
class/kernel: Remove those .*.cmd files

esben:
I don't believe there are any use of them in kernel-dev packages, and
many of them hardcode path to the originating recipe workdir, which
is obviously bad (with or without rmwork!).

Could/should we remove the .o files also?  Do they serve a purpose in
kernel-dev packages?

### DIFF
--- a/classes/kernel.oeclass
+++ b/classes/kernel.oeclass
@@ -365,7 +365,7 @@ do_install_kernel () {
         mkdir -p ${D}${kernelsrcdir}/$machdir
         cp -fR $d ${D}${kernelsrcdir}/$machdir/
     done
-    #find ${D}${kernelsrcdir}/ \( -name \*.o -o -name .\*.cmd \) -exec rm {} \;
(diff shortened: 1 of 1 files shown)


ravi-prevas:
We should probably try to make kernel-dev contain roughly what a distro's kernel-headers package contains, so certainly .o and .o.cmd do not belong there.

esben:
Ok, I updated with .o removal also.

ravi-prevas:
Pet peeve: Could you please not spawn a process for every file to delete; replace `-exec rm {} \;` by just `-delete`. Other than that, ack.

esben:
Done.

I was not aware of `-delete`, so I learned something today :)

KimBoendergaard:
Have just identified a project (ethercat-modules on powerpc) expecting the *.o files to be available. Or more precisely - I think the linker expects it.
Also found this thread https://bugs.launchpad.net/ubuntu/+source/linux/+bug/355344

PR #231, opened 2017-08-25
qemu and class for build-time ld.so.cache creation

Villemoes:
I haven't tested this much, but it does end up building a usable root file system.

### DIFF
--- a/recipes/qemu/qemu.inc
+++ b/recipes/qemu/qemu.inc
@@ -1,0 +1,52 @@
+SUMMARY = "QEMU is a generic and open source machine emulator and virtualizer."
+LICENSE = "GPL-2.0"
+
+SRC_URI = "https://download.qemu.org/qemu-${PV}.tar.xz"
+RECIPE_TYPES = "native"
+
+inherit c make pkgconfig
+
+DEPS = "libglib libgthread"
+
+DEPENDS += "${DEPS}"
+DEPENDS_${PN} += "${DEPS}"
+RDEPENDS_${PN} += "${DEPS}"
+
+EXTRA_OECONF = "--prefix=${prefix}"
+EXTRA_OECONF += "--disable-guest-agent-msi"
+EXTRA_OECONF += "--extra-cflags='${CFLAGS}'"
+
+DEPS += "libbz2"
+EXTRA_OECONF += "--enable-bzip2"
+
+DEPS += "liblzo2"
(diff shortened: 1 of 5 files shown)


esben:
Where do the root file system come from?
I would have guessed that this only would build the actual emulator(s).

Do this really build all the possible emulators?
And if so, isn't that rather time consuming?
And if so, what about having separate recipes for each architecture?

recipes/qemu/qemu.inc
```diff
@@ -0,0 +1,86 @@
...
+qemu-system-sparc64    \
+qemu-system-tricore    \
+qemu-system-unicore32    \
+qemu-system-x86_64    \
+qemu-system-xtensa    \
+qemu-system-xtensaeb    \
+qemu-tilegx    \
+qemu-x86_64    \
+"
+
```
esben: 2 x random trailing blank lines.

Villemoes:
This does just add the actual emulators. See the follow-up commit for an example where one provides the root filesystem for the emulation to use (and modify, which is the whole point). The "does end up building a usable root filesystem" comment was about that patch generating an ld.so.cache which is actually valid (in fact, byte-identical to running ldconfig on the target).

Yes, it does build all the emulators, and it does take some time. But I guess it will mostly be a one-time thing per manifest. I suppose one could create separate recipes for each emulation target, but I don't know how much time is done building generic stuff and how much is target specific.

Villemoes:
OK, so configuring qemu to just build the arm emulators takes 1.5 minutes to compile, as opposed to 7.5 minutes for the full monty. However, creating separate recipes for each arch ended up being rather messy, so I've tried a different route: let the manifest choose the arch's to support via USE_qemu_* flags. Then if somebody wants to use image_ldconfig.oeclass, he also needs to ensure that the appropriate flag is set (probably in his distro conf) so that there is a provider for the qemu-user-foo he needs.

esben:
LGTM. Can you edit the title so we can merge it?  Or do you want to add more stuff before merging?

Villemoes:
No, I don't currently have anything more lined up, though an image_bbdepmod.oeclass should now be easy to do if one wants to use the busybox MODPROBE_SMALL option and create its custom modules.dep.bb file at build time (real depmod, as used by the image_depmod.oeclass, works fine in a cross-compile setting, but generates files which are understood by either real kmod utils, or busybox without the MODPROBE_SMALL option). I'll change the title.

PR #233, opened 2017-09-20
class/elfwrapper: Do not fork

esben:
Avoid forking of, making a bit more convenient to keep track of the PID
of the process.


esben:
Would `-a $0` make sense?

Villemoes:
No idea what -a arg would make sense, I'm just worried whether the change to `exec` might change what argv[0] the program gets in a way that would break it, but I haven't really thought it through. One would probably have to do some experiments to see what happens. But if it doesn't break the sdk, it's probably fine, so feel free to merge it as-is.

esben:
I don't think that the argv[0] as seen by the program is changed with this change.

In both cases (before and after this PR),  argv[0] is the full path to the wrapped command.
Adding `-a $0` to exec would change argv[0] to be the name (or path) to the wrapper.
I believe that would make more sense, but is another change IMHO.

PR #234, opened 2017-09-28
class/elfwrapper: Set argv[0] to wrapper instead of wrapped cmd

esben:
When calling the wrapped command, set argv[0] to the wrapper
argv[0] instead of the full path to the wrapped command.

This should be safer, as argv[0] can be used by the wrapped
command (where as the full path to the wrapped command is not
safe to call directly).  Also, commands using argv[0] for usage
information looks nicer.


Villemoes:
OK if the sdk actually works; I've never really used an sdk generated by oe-lite.

esben:
It works :)

esben:
Actually, this does not work in cases where /bin/sh is not bash.
And as ld.so is shifting out argv[0] anyway, there really is no point in doing this anyway.

PR #235, opened 2017-10-03
class/autotools: Fix corner-case for CONFIGURE_SCRIPT variable

esben:
If (when) a recipe is machine specific, the weak assignment to
CONFIGURE_SCRIPT is causing problems.  It ends up assigning the
generic (not machine specific) WORKDIR value instead of the correct
machine specific variant, which is created by a hook later on.

This commit works around it, using an extra variable. *sigh*


Villemoes:
Huh? The RHS of `?=` is not expanded, so why doesn't a later setting of ${S}/${WORKDIR} work as expected?

esben:
Good question. It looked VERY much like RHS of `?=` is being expanded.

Villemoes:
Even if it were, I cannot possibly see how adding an intermediate variable would or could solve it.

Maybe some corner case in cache handling where we don't catch a machine-dependency? Or did you do a partial build with -t unpack or something like that first? I think I've seen EXTRA_ARCH break in such cases.

esben:
Argh, this is definitely not a proper fix to anything.

The problem I have stumbled upon is EXTRA_ARCH management when using MACHINE_* in variable overrides.  That turns out to basically not be working as of now.  So either avoid doing it at all, or make sure that EXTRA_ARCH is set by other means (fx. `EXTRA_ARCH="${MACHINE}"`).

As it is now, the machine override, when processed, will trigger setting of EXTRA_ARCH.  However, at that point in time, several variables (incl. functions) might have been expanded with what then turns out to be a wrong value for EXTRA_ARCH.  And making slight changes to metadata, can change the order of variable expansion, so the problem might disappear (or change) in unexpected ways.  :-1:

PR #243, opened 2018-03-12
Kernel compression

esben:


Villemoes:
Oh well, it's easy enough to do a replace-all-tabs-with-8-spaces afterwards if we (you) care enough about this, so I won't let that get in the way of merging as-is.

esben:
Ah, yes.  Sorry about that.  Bad Emacs!

PR #245, opened 2018-08-27
Linux headers 4.14

esben:


esben:
This should fix the libgpiod PR.

Villemoes:
OK, no reason to hold up this just because of some unrelated fetch issues.

PR #246, opened 2018-12-21
Add support for the meson build system, used with ex. gnome packages.

kjeldflarup:
Using meson requires installing ninja as well as python > 3.5.  
meson it self must also be installed via pip.

Disclaimer, this is only tested with arm target and with the gssdp package, see PR for that: https://github.com/oe-lite/base/pull/594

Signed-off-by: Kjeld Flarup <kfa@deif.com>

### DIFF
--- a/classes/meson.oeclass
+++ b/classes/meson.oeclass
@@ -1,0 +1,130 @@
+# build rules for meson
+# Needed for latest versions of gnome packages. 
+
+# remember to install
+# python >= 3.5
+# pip3 install meson
+# ninja-build
+
+inherit pkgconfig c c++
+
+def meson_array(d, var):
+    items = d.get(var).split()
+    return repr(items[0] if len(items) == 1 else items)
+
+# Map our ARCH values to what Meson expects:
+# http://mesonbuild.com/Reference-tables.html#cpu-families
+def meson_cpu_family(d, var):
+    import re
+    arch = d.getVar(var)
+    if arch == 'powerpc':
+        return 'ppc'
+    elif arch == 'powerpc64':
+        return 'ppc64'
+    elif arch == 'mipsel':
+        return 'mips'
+    elif arch == 'mips64el':
+        return 'mips64'
+    elif re.match(r"i[3-6]86", arch):
+        return "x86"
+    else:
+        return arch
+
+def meson_endian(d, prefix):
+    endian = d.getVar(prefix + "_ENDIAN")
+    if endian == 'l':
+        return "little"
+    elif endian == 'b':
+        return "big"
+    else:
+        bb.fatal("Cannot determine endianism for %s" % (prefix))
+
+B = "${WORKDIR}/build"
+# meson must run in the source dir
+do_configure[dirs] = "${B} ${S}"
+do_configure[cleandirs] = "${B}"
+
+OE_MESON_CROSSFILE = "${B}/meson.cross"
+OE_MESON_CONF = "--cross-file ${OE_MESON_CROSSFILE}" 
+EXTRA_OEMESON = ""
+MESON_TOOLCHAIN_ARGS = ""
+MESON_LINK_ARGS = "${@meson_array(d, 'LDFLAGS')}"
+MESON_LINK_ARGS[import] = "meson_array"
+MESON_C_ARGS = "${@meson_array(d, 'CFLAGS')}"
+MESON_C_ARGS[import] = "meson_array"
+MESON_CPP_ARGS = "${MESON_C_ARGS}"
+
+MESON_HOST_CPU_FAMILY = "${@meson_cpu_family(d, 'HOST_CPU')}"
+MESON_HOST_CPU_FAMILY[import] = "meson_cpu_family"
+MESON_TARGET_CPU_FAMILY = "${@meson_cpu_family(d, 'HOST_CPU')}"
+MESON_TARGET_CPU_FAMILY[import] = "meson_cpu_family"
+
+MESON_HOST_ENDIAN = "${@meson_endian(d, 'HOST')}"
+MESON_HOST_ENDIAN[import] = "meson_endian"
+MESON_TARGET_ENDIAN = "${@meson_endian(d, 'TARGET')}"
+MESON_TARGET_ENDIAN[import] = "meson_endian"
+
+override_native_tools() {
+    # Set these so that meson uses the native tools for its build sanity tests,
+    # which require executables to be runnable. The cross file will still
+    # override these for the target build.
+    export CC="${BUILD_CC}"
+    export CXX="${BUILD_CPP}"
+    export LD="${BUILD_LD}"
+    export AR="${BUILD_AR}"
+    export CPPFLAGS="${BUILD_CPPFLAGS}"
+    export CFLAGS="${BUILD_CFLAGS}"
+    export CXXFLAGS="${BUILD_CXXFLAGS}"
+    export LDFLAGS="${BUILD_LDFLAGS}"
+}
+
+do_configure() {
+	# generate the cross file, which meson uses to determine target compile options
+	cat > ${OE_MESON_CROSSFILE} <<EOF
+[binaries]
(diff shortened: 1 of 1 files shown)


classes/meson.oeclass
```diff
@@ -0,0 +1,123 @@
+# build rules for meson
+# Needed for latest versions of gnome packages. 
+
+# remember to install
+# python >= 3.5
+# pip3 install meson
+# ninja-build
+
+def meson_array(d,var):
```
esben: Please add space after comma (goes for entire file).

classes/meson.oeclass
```diff
@@ -0,0 +1,123 @@
...
+def meson_endian(d,prefix):
+    endian = d.getVar(prefix + "_ENDIAN")
+    if endian == 'l':
+        return "little"
+    elif endian == 'b':
+        return "big"
+    else:
+        bb.fatal("Cannot determine endianism for %s" % (prefix))
+
+OE_MESON_CROSSFILE = "${STAGE_DIR}/cross/meson.cross"
```
esben: This is bad.  You are not supposed to write to `${STAGE_DIR}`.  Please write this to `${B}` instead.

classes/meson.oeclass
```diff
@@ -0,0 +1,123 @@
...
+    # which require executables to be runnable. The cross file will still
+    # override these for the target build.
+    export CC="${BUILD_CC}"
+    export CXX="${BUILD_CPP}"
+    export LD="${BUILD_LD}"
+    export AR="${BUILD_AR}"
+    # These contain *target* flags but will be used as *native* flags.  The
+    # correct native flags will be passed via -Dc_args and so on, unset them so
+    # they don't interfere with tools invoked by Meson (such as g-ir-scanner)
+    unset CPPFLAGS CFLAGS CXXFLAGS LDFLAGS
```
esben: There are ${BUILD_CPPFLAGS} which is the build/native version of ${CPPFLAGS}. And same for the other flags.
Could you find a way to make meson use them for native instead of just clearing them?

classes/meson.oeclass
```diff
@@ -0,0 +1,123 @@
...
+	cat > ${OE_MESON_CROSSFILE} <<EOF
+[binaries]
+c = '$CC'
+cpp = '$CPP'
+ar = '$AR'
+nm = '$NM'
+ld = '$LD'
+readelf = '$READELF'
+strip = '$STRIP'
+pkgconfig = 'pkg-config'
```
esben: Shouldn't you add `inherit pkgconfig` in this class, so you get working cross pkg-config behavior?

classes/meson.oeclass
```diff
@@ -0,0 +1,123 @@
...
+
+	override_native_tools
+	# --prefix=/usr --libdir=lib looks like default values, but without them libdir will be locked to x86
+	# https://github.com/mesonbuild/meson/issues/2535
+	meson --prefix=/usr --libdir=lib builddir ${EXTRA_OEMESON} ${OE_MESON_CONF}
+}
+
+do_compile() {
+	override_native_tools
+	ninja -C builddir
```
esben: You should use `${B}` instead of hardcoded builddir.

kjeldflarup: Changed to ${B}/builddir. This is the same as the do_compile is executed in ${B}  
Meson requires buildir to be something else than ${B}, as this is seen as the source dir.

classes/meson.oeclass
kjeldflarup: Added, but the pkgconfig might not be used in my case, or meson handled pkgconfig otherwise.

classes/meson.oeclass
kjeldflarup: Changed to this but effect is unknown as the the build succeeds in both cases.
```
export CPPFLAGS="${BUILD_CPPFLAGS}"
export CFLAGS="${BUILD_CFLAGS}"
export CXXFLAGS="${BUILD_CXXFLAGS}"
export LDFLAGS="${BUILD_LDFLAGS}"
```
An inherit of `c` and `c++` was needed

classes/meson.oeclass
```diff
@@ -0,0 +1,125 @@
...
+def meson_endian(d, prefix):
+    endian = d.getVar(prefix + "_ENDIAN")
+    if endian == 'l':
+        return "little"
+    elif endian == 'b':
+        return "big"
+    else:
+        bb.fatal("Cannot determine endianism for %s" % (prefix))
+
+OE_MESON_CROSSFILE = "${B}/meson.cross"
```
esben: Should should add
```
B = "${S}/builddir"
```
to avoid cluttering `${S}`

classes/meson.oeclass
esben: No, you probably just ended up using host pkg-config.  Which might, or might not be problematic, depending on in which context (native or cross) it is used.

classes/meson.oeclass
```diff
@@ -0,0 +1,125 @@
...
+c_args = ${MESON_C_ARGS}
+c_link_args = ${MESON_LINK_ARGS}
+cpp_args = ${MESON_CPP_ARGS}
+cpp_link_args = ${MESON_LINK_ARGS}
+EOF
+
+	override_native_tools
+	# --prefix=/usr --libdir=lib looks like default values, but without them libdir will be locked to x86
+	# https://github.com/mesonbuild/meson/issues/2535
+	meson --prefix=/usr --libdir=lib ${B}/builddir ${EXTRA_OEMESON} ${OE_MESON_CONF}
```
esben: Please use `${B}` instead of `${B}/builddir` with `B = "${S}/builddir"`.

classes/meson.oeclass
kjeldflarup: Negative. meson must execute in ${S}  
Specifying ${B}/builddir seems to be the meson way to avoid cluttering!

Villemoes: Yes, that's the proper way to do this. Also, please remember to add `do_configure[cleandirs] = "${B}"`. And can we set B = WORKDIR/build instead of making it a subdir of `${S}`? I'm worried some repo down the road might have a non-empty `builddir`, for whatever reason, so it seems better to stay completely out of `${S}`, no?

Villemoes: oh, sorry, you already do exactly that. Please ignore me.

classes/meson.oeclass
```diff
@@ -0,0 +1,130 @@
...
+def meson_endian(d, prefix):
+    endian = d.getVar(prefix + "_ENDIAN")
+    if endian == 'l':
+        return "little"
+    elif endian == 'b':
+        return "big"
+    else:
+        bb.fatal("Cannot determine endianism for %s" % (prefix))
+
+B = "${WORKDIR}/builddir"
```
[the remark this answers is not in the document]
kjeldflarup: Builddir was chosen because meson uses `builddir` in all examples. But that was of course when creating builddir as a subdirectory of the source.

classes/meson.oeclass
kjeldflarup: The gssdp package is using pkg-config, without a working pkg-config it cannot find glib-2.0, thus it seems fair to assume that this works for cross if pkg-config found a native library, build would assumably fail.  
Native build is not needed for the current meson use, thus it is not testet.

classes/meson.oeclass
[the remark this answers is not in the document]
esben: Yes, and it was fine in that context.  But please change it to build now that it is a subdir of WORKDIR.

classes/meson.oeclass
esben: Ok, let's leave it like this for now.

classes/meson.oeclass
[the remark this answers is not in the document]
kjeldflarup: done

esben:
@kjeldflarup You have messed up the PR. There are now 16 commits...

kjeldflarup:
@esben fixed

